PR #132, opened 2025-05-12
Begin cleanup of project and update README based on onboarding

danielfay:
The PR starts to chip away at getting this project setup for a new dev.

The overall goal has been to make the README and surrounding project behave as a new comer would expect.

- The project now points to the Amplfy staging env
- All package scripts now work and pass
- Tests that became invalid during the 2023 rework have been removed
- CircleCI has be removed in favor or using Github Actions.


test/karma/transcribe/spec/store.region.spec.js
danielfay, line 1: The two tests below are currently skipped, but will be 'fixed' in my next PR. They are running, but failing and I think something has changed with how things are saved.

Didn't make sense to remove them since they reflect the code, just are broken right now.

aaronfay (approved):
Some tests removed that I _feel_ like we might need back at some point. But overall good to go 👍🏻

PR #133, opened 2025-05-22
Protect regions from being deleted by anyone but the author

danielfay:
Disables the ability to delete regions if you are not the transcription author (even if you are an editor).

There is currently no way to delete transcriptions, but it will likely be locked down the same way.


aaronfay (approved):
Looks good, thanks for this 🙏🏻 save us at least a couple headaches

PR #147, opened 2025-07-08
136 add lambda devprocessmedia to infra model

### LINKED ISSUE #136: Add Lambda `devProcessMedia` to infra model
When you upload a new transcription, the _nominal_ process is: 

 * DynamoDB trigger kicks off Lamnda function `devProcessMedia`
 * this function pulls down the `.mp3`/`.mp4` (and extracts audio if necessary)
 * special sauce calculations happen to translate the waveform into JSON (for wavesurfer)
 * the JSON file is written back to S3 under <bucket>/public/<original-file-and-extension>.json

There are a couple main issues with this:

 * the Lambda is hand-rolled, I'm not even sure the code is in any repo
 * Lambda is not managed by any IaC
 * the Lambda currently handles _any_ env you point at it

danielfay:
- Add peak file creation to amplify (doesn't include the EFS that must be created manually for now)
- Remove enqueue region api and wire lambda to region table
- Add queue needed for indexing to amplify
- Make index lambdas config dynamic

Yet to be tested, there may be some policy/permission changes to be made


aaronfay:
At a glance, looks amazing. I really think when you're feeling ready, let's merge this down and run it a few times against the new `staging` env, give it a test drive. The "missing bits" will poke out immediately I think.

PR #149, opened 2025-07-14
Migrate crk FST spell checking to amplify

### LINKED ISSUE #137: Lambda `dev-crk-inflect-worker` needs additional modelling
The Lambda `dev-crk-inflect-worker` is currently managed in https://github.com/kiyanaw/crk-lookup-api, but there is a Lambda layer that includes the bundled FST from https://github.com/giellalt/lang-crk. 

Additionally, I don't believe there are multiple environments, similar to `devProcessMedia` I believe one API is doing all the work for each environment.

## Tasks

 - model the Lambda layer with the FST
 - build out environments to properly support `production`, `staging` etc.

danielfay:
Currently getting `var/task/node_modules/hfstol/build/Release/hfstol_addon.node: invalid ELF header"` up in the lambda.

I may have to set up a linux virtual machine and see if pushing it from there will match the lambda architecture. Very strange.


aaronfay:
Nice work, take your time with it. This will allow us to play with the new FTSs at some point and see how they work, looking forward to it.

aaronfay (approved):

PR #155, opened 2025-07-29
Elasticsearch setup

### LINKED ISSUE #139: Model ElasticSearch setup
Right now ES is hand-rolled and is _super super_ old. We need a couple processes:

 * ability to blow away and rebuilt ES from scratch
 * ability to re-index all data from Dynamo

danielfay:
The PR creates the OpenSearch domain using a custom cloudformation. It also fixes some bugs in the related indexing lambdas.

Indexing is still broken because the region dynamodb table is different now. That'll be the next thing I fix.

Also, `IsProduction` is my attempt to future proof these for the production AWS account. We'll see how it goes.

## Notable Parts
- OpenSearch and Cloudformation interact a little strangely  eg. If you had settings and then remove them from the Cloudformation unless they explicitly changed it will leave the OpenSearch configured as is.
- Related to the above while there is a `MasterUserPassword` in the Cloudformation it is only used as a temp password when the OpenSearch is first created. I've already updated it and we can put it in a shared 1password or something.
- OpenSearch granular permissions are done through security users and roles managed in the OpenSearch dashboard.
- Did you know that in the new lambda run time environments `aws-sdk` is no longer including for you? While surprising I like this change aha.


aaronfay (approved):
LLCFTM "looks like cloudformation to me" 😂 

Nice work, cleaning up those lambdas feels pretty good 👍🏻

PR #156, opened 2025-07-31
Adds ability to invite users to collaborate on transcription

aaronfay:
This change introduces some new infra, as well as a couple screens to manage the sharing of transcriptions. 

## Infra
 * New schema model for Invitations  (Invite)
 * SES for sending emails
 * Lambda (invite) for handling of sending, listing, revoking invitations
 * Removes stale model "Contributor" + related references

## UI
 * Manage Invitations via Transcription settings page
 * Revoke sent Invitations
 * Invitations menu item with a list of all invitations & their statuses
 * Accept invitation dialog
 * Adjusts transcription list to show which transcriptions have been shared (author) and as collaborator what permissions we have

### Share settings
screenshot-2025-07-31-13 18 04@2x

### Invite email
screenshot-2025-07-31-13 16 52@2x

### Invite view
screenshot-2025-07-31-13 17 16@2x

### Accept invite
screenshot-2025-07-31-13 17 25@2x

### Sharing icons
screenshot-2025-07-31-13 24 31@2x
screenshot-2025-07-31-13 23 41@2x


amplify/backend/api/invite/cli-inputs.json
danielfay, line 1: A discussion for a different time, but I may change this to a generic kiyanaw rest api when we get around to adding transcription deletes. For now this looks great 😄

amplify/backend/api/kiyanaw/schema.graphql
```diff
@@ -15,7 +15,8 @@ type Transcription
     { allow: groups, groups: ["Admins"], operations: [create, read, update, delete] }
   ]) {
   id: ID! @primaryKey
-  author: String!
+  author: String! # UUID of the Cognito user
+  authorFriendly: String! # Email/friendly identifier
```
danielfay: This is handy for if we decide to use profiles or something instead.

I'm not 100% on board with using email all over the site for the author. Emails are PII so we should stay mindful.

amplify/backend/function/inviteHandler/src/actions/services/invite-service.js
```diff
@@ -0,0 +1,212 @@
...
+      // Add required fields for Amplify
+      __typename: 'Invite',
+      updatedAt: inviteData.createdAt,
+      // Add required DataStore fields
+      _version: 0,
+      _lastChangedAt: 0,
+      _deleted: false
+    };
+
+    const command = new PutCommand({
```
danielfay, line 60: I have no idea why they decided on the verb `Put` for the create command, but yeah, that's correct. Interesting.

amplify/backend/function/inviteHandler/src/actions/templates/invite-email.js
```diff
@@ -0,0 +1,57 @@
...
+ * @param {string} data.expiryDate - Human-readable expiry date
+ * @returns {Object} Email content with subject, htmlBody, textBody
+ */
+function generateInviteEmail(data) {
+  const { invitedBy, transcriptionTitle, permissionLevel, inviteLink, expiryDate } = data;
+
+  const action = permissionLevel === 'viewer' ? 'view' : 'edit';
+  const subject = `Invitation to ${action} transcription "${transcriptionTitle}"`;
+
+  const htmlBody = `
```
danielfay, line 17: I'm alright with this for now, but we should on how we're going to consolidate branding on things like this.

We can probably have main.css in a lambda layer for example so that lambdas that need to send emails or notifications have easy access to colours and logos and stuff (and they can be changed in one place and update all the emails). And they they can just spit it into the top of the `htmlBody` or something.

Not worried about it for today though.

amplify/backend/function/inviteHandler/src/actions/use-cases/accept-invite.js
```diff
@@ -0,0 +1,240 @@
...
+        Key: {
+          id: transcriptionId
+        },
+        UpdateExpression: `SET #list = list_append(if_not_exists(#list, :empty_list), :user_id), #updatedAt = :updatedAt`,
+        ExpressionAttributeNames: {
+          '#list': listAttribute,
+          '#updatedAt': 'dateLastUpdated'
+        },
+        ExpressionAttributeValues: {
+          ':user_id': [userId],
```
danielfay, line 163: It wasn't obvious to me why this was a list.

It seems list_append merges two lists. ie. `list_append(list1, list2)`

amplify/backend/function/inviteHandler/src/actions/use-cases/revoke-invite.js
```diff
@@ -0,0 +1,165 @@
...
+    if (invite.invitedBy !== requestorUserId) {
+      console.error(`❌ Authorization failed: ${invite.invitedBy} !== ${requestorUserId}`);
+      throw new Error(`Unauthorized: Only the person who sent the invite can revoke it. Invite was sent by "${invite.invitedBy}" but revoke requested by "${requestorUserId}"`);
+    }
+
+    // 3. If the invite was accepted, remove the user from the transcription ACLs
+    if (invite.status === 'accepted') {
+      console.log(`📝 Invite was accepted - removing user from transcription ACLs`);
+      
+      // Use acceptedByUserId if available (new field), fallback to email for older invites
```
danielfay, line 136: Is this fallback needed? If you delete the invites made during your development, there is no such thing as 'older invites' right?

amplify/backend/function/inviteHandler/src/actions/use-cases/send-invite.js
```diff
@@ -0,0 +1,154 @@
...
+    const emailContent = generateInviteEmail({
+      invitedBy: invitedByFriendly,
+      transcriptionTitle,
+      permissionLevel,
+      inviteLink,
+      expiryDate,
+    });
+
+    // Send email via service
+    const result = await emailService.sendEmail({
```
danielfay: If this email fails to send (mostly likely if it bounces because of a typo or something), we should probably catch it and update the invite status from 'pending' to some kind of error status so the user knows it didn't work (beyond just seeing the toast of what have you.

amplify/backend/function/inviteHandler/src/index.js
```diff
@@ -0,0 +1,222 @@
...
+                    body: JSON.stringify({
+                        success: true,
+                        ...result
+                    })
+                };
+            } catch (error) {
+                console.error('Invite handler error:', error);
+                
+                // Determine if it's a validation error (400) or server error (500)
+                const isValidationError = error.message.includes('Missing required parameters') ||
```
danielfay: This type of thing strikes me as super fragile. Maybe preface the validation error messages with 'Validation Error' so that you aren't checking the specific messages. Those messages may get updated in the future for clarity.

amplify/backend/function/inviteHandler/src/index.js
```diff
@@ -0,0 +1,222 @@
...
+
+const { handleInvite } = require('./actions/handle-invite');
+const { handleAcceptInvite } = require('./actions/handle-accept-invite');
+const { handleGetMyInvites } = require('./actions/handle-get-my-invites');
+const { handleRevokeInvite } = require('./actions/handle-revoke-invite');
+
+/**
+ * API Gateway Lambda router for handling multiple endpoints
+ * 
+ * Supported routes:
```
danielfay, line 51: Revoke also seems to be supported.

amplify/backend/function/inviteHandler/inviteHandler-cloudformation-template.json
```diff
@@ -0,0 +1,444 @@
+{
+  "AWSTemplateFormatVersion": "2010-09-09",
+  "Description": "{\"createdOn\":\"Mac\",\"createdBy\":\"Amplify\",\"createdWith\":\"14.0.0\",\"stackType\":\"function-Lambda\",\"metadata\":{\"whyContinueWithGen1\":\"\"}}",
+  "Parameters": {
+    "CloudWatchRule": {
```
danielfay, line 5: This was probably added by mistake. I'm not sure if removing it will make another part of Amplify angry though. Maybe be best just to leave it if everything is working.

src/components/auth/AuthenticatorShell.tsx
```diff
@@ -20,11 +20,13 @@ const AuthenticatedApp = ({ children }: { children: ReactNode }) => {
   if (lastUserRef.current !== currentUserKey) {
     lastUserRef.current = currentUserKey;
     
-          const currentUser = route === 'authenticated' && user ? {
-        username: user.signInDetails?.loginId as string,  // This is the email
-        userId: user.userId,
-        signInDetails: user.signInDetails,
-      } : null;
+    const currentUser = route === 'authenticated' && user ? {
+      username: user.signInDetails?.loginId as string,  // This is the email
```
danielfay, line 24: I would like us to eventually decide on something else for username.

One of the things we were planning on changing on the old site was that when you sent and invite you had a list of the member emails.

With username being the email, the public transcription list will basically be the same thing all over again. Just on the different page 😛

src/components/forms/TranscriptionSettingsPage.tsx
```diff
@@ -9,8 +13,9 @@ interface TranscriptionSettingsPageProps {
   author: string;
   dateLastUpdated: string;
   regionCount: number;
+  transcriptionId: string; // Added this prop
```
danielfay: A couple benign comments in this file.

src/components/forms/TranscriptionSettingsPage.tsx
```diff
@@ -21,11 +26,50 @@ export const TranscriptionSettingsPage = ({
...
+
+  const loadInvites = useCallback(async () => {
+    setInvitesLoading(true);
+    
+    try {
+      if (!transcriptionId?.trim()) {
+        throw new Error('Transcription ID is required');
+      }
+      console.log('Loading invites for transcription:', transcriptionId);
+      const loadedInvites = await services.inviteService.loadInvitesForTranscription(transcriptionId);
```
danielfay: I believe this is the only use of `services` in this file. Seems worthwhile to just import `inviteService` instead of the whole services folder generically up top.

It would also more clearly show what actual services are used in the settings.

src/components/forms/TranscriptionSettingsPage.tsx
```diff
@@ -21,11 +26,50 @@ export const TranscriptionSettingsPage = ({
...
+  
+  // Hooks
+  const createInvite = useCreateInvite();
+  const revokeInvite = useRevokeInvite();
+
+  const loadInvites = useCallback(async () => {
+    setInvitesLoading(true);
+    
+    try {
+      if (!transcriptionId?.trim()) {
```
danielfay: This is probably all over and it's the first time I'm noticing, but do you have any idea why these are being `trim`'d? They aren't user entered, why would they ever have extra white space?

src/components/forms/TranscriptionSettingsPage.tsx
```diff
@@ -64,6 +108,87 @@ export const TranscriptionSettingsPage = ({
 
   const hasChanges = title !== initialTitle || comments !== (initialComments || '');
 
+  const handleSendInvite = async () => {
+    if (!newInviteEmail.trim()) {
```
danielfay: In contrast I get this. It was typed by a user.

Though I do notice two more `trim`s below. I would prefer that the value was cleaned once and could be trusted in the rest of the code.

src/components/list/TranscriptionsTable.tsx
```diff
@@ -23,23 +26,23 @@ const columns: Column[] = [
   { key: 'issues', label: 'Issues', width: '8%', sortable: true },
   { key: 'dateLastUpdated', label: 'Last Edit', width: '15%', sortable: true },
   { key: 'type', label: 'Type', width: '8%', sortable: true },
-  { key: 'source', label: 'Source', width: '8%' },
+  // { key: 'source', label: 'Source', width: '8%' },
```
danielfay: Were you meaning to push this back? If not you may as well remove it.

Maybe write a ticket to figure out how we're handling transcription sources instead?

src/hooks/useLoadInvite.ts
```diff
@@ -0,0 +1,31 @@
...
+export interface InviteValidation {
+  isValid: boolean;
+  isExpired: boolean;
+  isPending: boolean;
+  isAccepted: boolean;
+  canAccept: boolean;
+}
+
+/**
+ * Hook for accessing current invite store data - NO useEffect!
```
danielfay, line 12: A war took place here lol

src/hooks/useLoadTranscriptions.ts
```diff
@@ -0,0 +1,28 @@
...
+    store.setLoading(true);
+    
+    const useCase = new LoadTranscriptions({
+      services,
+      store,
+    });
+
+    useCase.execute().catch((error) => {
+      console.error('❌ useLoadTranscriptions failed:', error);
+      // Error is already handled in the use-case
```
danielfay: What does 'handled' mean in this context? If it was handled in the use-case why does it need to be caught here?

It's more likely to me that an unexpected error in the use-case gets caught in the hook.

src/hooks/useRevokeInvite.ts
```diff
@@ -0,0 +1,32 @@
...
+
+  const revokeInvite = useCallback(async (inviteId: string): Promise<RevokeInviteResult> => {
+    if (!user) {
+      throw new Error('User must be authenticated to revoke invitations');
+    }
+
+    const revokeInviteUseCase = new RevokeInviteUseCase({
+      inviteId,
+      requestorUserId: user.userId,
+      services
```
danielfay, line 25: I got a bit curious about some of the use-cases passing in all the services. In this case looking at the revoke invite use-case it only seems to use the `inviteService`.

I think it is worthwhile to be more explicit about what services a use-case actually relies on. It also encourages writing smaller use-cases that only rely on a couple services.

src/pages/InvitationsPage.tsx
danielfay, line 1: I'm coming to the realization that if we ever end up tackling internationalization it's going to be a bit of a nightmare aha. 

Related, all of the 100% custom markup in these tsx files is going to make theming for different groups in the future hairy as well. We should sit down and think about it when we're in a more stable place.

src/services/inviteService.ts
```diff
@@ -0,0 +1,665 @@
+import { generateClient } from 'aws-amplify/api';
+import { post } from 'aws-amplify/api';
```
danielfay: I'm surprised this ended up on two lines.

src/services/regionService.ts
danielfay, line 1: Impressive refactor. Well done!

src/use-cases/load-transcription.ts
```diff
@@ -30,6 +30,8 @@ export class LoadTranscription {
     const userService = this.config.services.userService
     
     const data = await transcriptionService.loadInFull(this.config.transcriptionId);
+
+    console.log('load in full', data)
```
danielfay: Probably don't need this.

package.json
```diff
@@ -13,7 +13,8 @@
     "format": "prettier --write src/",
     "preview": "vite preview",
     "test": "jest --silent",
-    "test:watch": "jest --watch"
+    "test:watch": "jest --watch",
+    "test:lambda": "cd amplify/backend/function/inviteHandler/src && npm install && npm test"
```
danielfay, line 17: When I (or you) end up adding tests to the second lambda this and the github action will need to be figured out. For now I think this is fine.

danielfay (approved):
We can double back on the bigger discussions in my comments. If this is all working as is, I have no issue merging if after you've read through my comments.

amplify/backend/api/kiyanaw/schema.graphql
aaronfay: It should be ok for now. The only folks who see this are someone you are sharing the transcription with, we're not plastering it all over the place. It definitely will be good to have a "username" option once we have user profiles, for now I'm just showing the first bit of the email, eg: `aaron.j.fay`

aaronfay: Another way I could approach this is to _just_ store the first bit of the email for now, and when we have profiles use the username 🤔 I'll make that change today. Easy to do

amplify/backend/function/inviteHandler/src/actions/templates/invite-email.js
aaronfay: Yup, good call.

amplify/backend/function/inviteHandler/src/actions/use-cases/accept-invite.js
aaronfay: Yeah, both `viewers` and `editors` are lists, so I assume this is how it works 🤷🏻

amplify/backend/api/kiyanaw/schema.graphql
danielfay: That works for me. I noticed later in the PR that you had some helpers that was already doing that before displaying the email. With that in mind it's probably for the best to just store it that way anyway.

package.json
aaronfay: Yeah, they were originally in the Jest config and just ran with the main `npm test` run, but that exploded in GitHub from missing deps. I did prefer to have them in the main test run, perhaps I just need a step that does the install for each function (if needed) before the test run. 🤔

danielfay: Indeed, I'm not sure what the best approach is. I did a little reading and jest also has the concept of 'projects', but I still don't think that will help with the npm installs needing to be run for each of the functions.

As I said, not losing sleep over it. When tests get added to another one we'll change it then.

amplify/backend/api/kiyanaw/schema.graphql
aaronfay: I've modified this to only use the first part of the email for now, in anticipation of "proper" profiles/usernames.

amplify/backend/function/inviteHandler/src/actions/use-cases/send-invite.js
aaronfay: Fixed, will set the invite to "failed" if the email fails.

amplify/backend/function/inviteHandler/src/index.js
aaronfay: Good call, I've cleaned this up using custom error types.

amplify/backend/function/inviteHandler/inviteHandler-cloudformation-template.json
aaronfay: I believe that was when it asked me if I wanted it on a schedule? Added automatically in any case.

src/components/auth/AuthenticatorShell.tsx
aaronfay: So for "recording" this value anywhere, I'm only using the first part of the email now (eg: `aaron.j.fay`) and once we have profiles we can let the user pick a username. For now this value only is shown to the actual user.

src/components/forms/TranscriptionSettingsPage.tsx
aaronfay: Not sure. Claude needs a trim today I guess 🤷🏻

src/components/list/TranscriptionsTable.tsx
aaronfay: I'm going to refactor this whole page, but pulled it out for now.

src/hooks/useLoadInvite.ts
aaronfay: I'm happy to leave those little nuggets in :)

src/hooks/useLoadTranscriptions.ts
aaronfay: Fixed.

src/hooks/useRevokeInvite.ts
aaronfay: Yeah, it's a bit of a "loophole" in the pattern, mainly it was intended to allow for simpler testing by injecting services. I've started writing some use-cases that just import services at the top of the file, and it doesn't seem to impede anything. I'll think about it some more.

src/pages/InvitationsPage.tsx
aaronfay: I don't have theming in mind for this app, only the mobile app, and hopefully only *once* haha. For i8ln tho you're right. We'll cross that bridge when we get there.

src/services/inviteService.ts
aaronfay: Right?

package.json
aaronfay: I've just re-consolidated the test runs, which I prefer. We'll need to remember to install deps for any lambda functions we want to unit test, but the tests should all run together now.

amplify/backend/api/invite/cli-inputs.json
aaronfay: Let's discuss this at some point, I may have missed a step here or chosen the wrong option?

PR #157, opened 2025-08-02
Updates transcription on region save

aaronfay:
Couple fixes in this PR:

 * when the transcription first loads and wavesurfer is "ready", save the transcription length
 * when adding/updating/deleting regions, update the transcription details (coverage, lastuser, etc)
 * makes consistent use of `store` in use cases

### DIFF
--- a/src/stores/useEditorStore.ts
+++ b/src/stores/useEditorStore.ts
@@ -50,9 +50,13 @@ interface EditorState {
   setSelectedRegion: (regionId: string | null) => void;
   setPlaybackWithinRegion: (regionId: string | null) => void;
   addNewRegion: (region: RegionData) => void;
+  deleteRegion: (regionId: string) => void;
   setRegionText: (regionId: string, text: string) => void;
   setRegionTranslation: (regionId: string, translation: string) => void;
   updateRegionBounds: (regionId: string, start: number, end: number) => void;
+
+  // Transcription metadata helpers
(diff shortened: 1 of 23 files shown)


src/stores/useEditorStore.ts
```diff
@@ -216,6 +220,31 @@ export const useEditorStore = create<EditorState>()(
...
+        // Remove from regions array
+        const newRegions = regions.filter(r => r.id !== regionId);
+        
+        // Prepare the update object
+        const updateObj: Partial<EditorState> = {
+          regions: newRegions,
+          regionMap: newRegionMap,
+        };
+        
+        // If this was the selected region, clear the selection
```
danielfay, line 239: No harm in the `if`, but is it possible to delete a region without selecting it? Maybe if you like, hit delete and then quickly click away to another region. 

Maybe keep the `if` just in case aha.

src/use-cases/update-region-text.ts
```diff
@@ -56,8 +56,7 @@ export class UpdateRegionTextUseCase {
     const timeoutKey = `region-text-save-${regionId}`;
     Timeout.set(timeoutKey, async () => {
       try {
-        // Get fresh store state at save time
-        const currentStoreState = store.getState();
+        // Get fresh state at save time
```
danielfay, line 59: I'm not sure what this comment is referencing. I think it should've been removed as well.

danielfay (approved):
Looks great, very minor notes.

Feel free to merge after looking at them 😃

PR #158, opened 2025-08-03
Adds realtime support + conflict resolution

aaronfay:
This change adds the necessary changes to support 

 * subscription to realtime region changes
 * visual indicator (slight green flash) that region has changed (both list + timeline)
 * auto-merge for non-conflicting fields (user A changes "text", user B changes position)
 * conflict dialog for conflicting changes

This is _not a perfect system_ by any stretch, if there are _three_ users working on the same transcription, _I have no idea what's going to happen_ 😂 but it's a good start. 

## Flash indicator of changes
https://github.com/user-attachments/assets/35a95546-ecef-44fe-8802-daa54ffbed24

## Conflict dialog

https://github.com/user-attachments/assets/f53d3146-e80c-48b4-b5f4-5b462d0b4856


src/services/conflictResolutionService.ts
```diff
@@ -0,0 +1,94 @@
...
+      // Use the global dialog manager to show the real UI dialog
+      const { conflictDialogManager } = await import('./conflictDialogManager');
+      const result = await conflictDialogManager.showDialog(conflict);
+      
+      console.log('🔥 User resolved conflict:', result);
+      this.activeConflicts.delete(conflictKey);
+      return result;
+    } catch (error) {
+      console.error('Error showing conflict dialog:', error);
+      // Fallback to accepting remote changes if dialog fails
```
danielfay: If I'm reading this right in `src/hooks/useConflictDialog.tsx`. Will this also trigger if the user cancels the dialog?

Since the conflict(s) need to be resolved to continue I think that accepting remote even when they close the dialog is probably the only way to go.

That said, I would find it pretty jarring if I cancelled a dialog box and the text I was working on changed.

src/services/flashIndicatorService.ts
```diff
@@ -0,0 +1,142 @@
...
+
+class FlashIndicatorServiceImpl {
+  private listeners = new Map<string, Set<FlashEventCallback>>();
+
+  /**
+   * Trigger a flash for a specific region
+   */
+  flashRegion(regionId: string, userEmail: string): void {
+    // Extract username from email (everything before @)
+    const username = userEmail.includes('@')
```
danielfay, line 41: When we have profiles there will be so many places where services sanitize there own username.

The saving grace may be that they all use `.includes('@')` so they will hopefully be easy to find.

src/services/versionConflictService.ts
```diff
@@ -0,0 +1,236 @@
...
+    // Show conflict resolution dialog
+    const resolution = await services.conflictResolutionService.showConflictDialog(conflictData);
+    
+    // Handle user's resolution choice
+    if (resolution.action === 'accept_remote') {
+      console.log('🔄 User chose to accept remote changes');
+      if (conflictAnalysis.remoteRegion) {
+        await onAcceptRemote(conflictAnalysis.remoteRegion);
+      } else {
+        console.error('Cannot accept remote changes: no remote region data available');
```
danielfay: Though, I'm not really sure how you would get in this state, if you call neither `onAcceptRemote` or `onKeepLocal` then what state does that leave the user in?

I suppose they will just get the dialog again when they make another edit and hopefully it works out that time.

src/use-cases/subscribe-to-region-changes.ts
```diff
@@ -0,0 +1,334 @@
...
+    
+    // Update version tracking
+    if (updatedRegion._version !== undefined) {
+      store.setRegionVersion(updatedRegion.id, updatedRegion._version);
+    }
+  }
+
+
+
+  // Note: Removed complex conflict prediction methods - we now use simple version-based conflicts
```
danielfay: Has this been done?

src/use-cases/subscribe-to-region-changes.ts
```diff
@@ -0,0 +1,334 @@
...
+      
+      // Update wavesurfer region bounds
+      wavesurferService.setRegionPosition(updatedRegion.id, {
+        start: updatedRegion.start,
+        end: updatedRegion.end
+      });
+    }
+
+    // Update region analysis (safe to update during any editing)
+    if (updatedRegion.regionAnalysis !== undefined) {
```
danielfay: There is some code duplication in this function and the one below that would probably help make this file feel less complicated.

Updating region analysis is done both here and on ln 318 for example.

src/use-cases/subscribe-to-region-changes.ts
```diff
@@ -0,0 +1,334 @@
...
+    // This ensures conflicts are detected when same field is edited, but allows parallel saves for different fields
+    
+        // BRILLIANT USER INSIGHT: Compare subscription vs BASELINE (before edits), not current store
+    // This tells us exactly what the OTHER browser changed
+    
+    // Get baseline from pending edits (original state when edit started)
+    const baseline = store.getBaselineForRegion(updatedRegion.id);
+    
+    // Compare subscription to baseline to see what ACTUALLY changed
+    const actualChanges = {
```
danielfay, line 227: This file is feeling very complicated to me and I've been trying to put my finger on why. I think part of it is that this method is a very large stream of consciousness.

The could be a method for example. Then at least the logic would be moved out of the way when you're trying to follow the remote saving itself.

Just below `hasUnprotectedChanges` is already perfectly named to have it's logic pushed a layer down.

src/use-cases/update-region.ts
```diff
@@ -0,0 +1,284 @@
...
+  translation?: string;
+  start?: number;
+  end?: number;
+  regionAnalysis?: string[];
+};
+
+interface UpdateRegionConfig {
+  regionId: string;
+  changes: RegionChanges; // e.g., { regionText: "new text" } or { start: 10, end: 20 }
+  debounceMs: number; // 3000 for text, 2500 for bounds
```
danielfay: Did you pick these numbers of a reason? It makes this seem fragile and tightly tuned. Can't they both just use the same default? (like, 3000?)

danielfay (approved):
Very neat stuff.

Some of the files under the hood are a bit unwieldy to follow. That can all be improved if the system gets worked on more. If it does the job and keeps trucking away then I'm happy to just leave it aha.

src/services/flashIndicatorService.ts
aaronfay: Yeah, I'm definitely cutting corners here, hopefully it won't be too painful.

src/services/conflictResolutionService.ts
aaronfay: Yeah, I've removed the cancel button and defaulted to "accept their changes". Good call. 

screenshot-2025-08-03-21 48 50@2x

src/services/versionConflictService.ts
aaronfay: Yeah, looks dumb. Claude does that a lot, unnecessary "defensive programming". I've removed it.

src/use-cases/subscribe-to-region-changes.ts
aaronfay: Claude leaves such dumb notes all over. Removed.

src/use-cases/subscribe-to-region-changes.ts
aaronfay: I cleaned up this file, it was refactored SO many times, it's definitely a mess. I've cleaned it up and added some explanation docstrings to some of the more complex bits.

src/use-cases/subscribe-to-region-changes.ts
aaronfay: Just refactored this and cleaned it up, I think it's better now, or at least less complex/more clear

src/use-cases/update-region.ts
aaronfay: I think they are actually both 2500, not sure where Claude got those numbers 😅

PR #161, opened 2025-08-07
A few fixes

aaronfay:
Finally, a small PR!

Mainly fixes where User B is viewing a conflict and User A makes another change, updates the dialog for User B.

### DIFF
--- a/src/services/conflictDetectionService.ts
+++ b/src/services/conflictDetectionService.ts
@@ -19,6 +19,35 @@ export interface ConflictResult {
 
 export class ConflictDetectionService {
   /**
+   * Helper function to normalize null/undefined/empty string values for comparison
+   * @param value The value to normalize
+   * @returns Normalized string value
+   */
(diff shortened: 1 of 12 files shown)


src/services/conflictDetectionService.ts
```diff
@@ -29,6 +29,14 @@ export class ConflictDetectionService {
       throw new Error('Cannot detect conflicts between different regions');
     }
 
+    // Helper function to normalize null/undefined/empty string values for comparison
+    const normalizeEmptyValue = (value: string | number | boolean | string[] | null | undefined): string => {
```
danielfay: This and the `hasConflict` logic below can likely be moved into class methods since they are used twice in this class.

danielfay (approved):
A nice bit of clean up in here, thanks!

One minor change noted.

src/services/conflictDetectionService.ts
aaronfay: Fixed.

PR #163, opened 2025-08-10
Fix region indexing

aaronfay:
Re-enables region indexing to OpenSearch.


danielfay (approved):
Awesome, looks great.

As I said in the call, I'm a bit surprised that you need to provide the `AWS.config` when making the OpenSearch client and that it doesn't look it's IAM role up itself since it is a lambda up in AWS. Oh well.

An interesting side effect of needing to do it that way is that it can be extended to allow for local execution if that is something we wanted to do in the future.

danielfay (approved):

PR #165, opened 2025-08-10
Transcription mass reindex

aaronfay:
Adds ability to delete transcriptions (with adequate warning) and infra to support it. Also handles mass-reindexing of data, regions/issues if various toggles are twiddled on a transcription.

Specifically: 

 * adds `publicIssues` property to Transcription model and new toggle in the UI,
 * adds issue indexing to `issues-<env>` for the OpenSearch index,
 * adds new function `onTranscriptionChange` to handle deleting transcription, mass re-indexing regions and issues, mass deleting regions and issues (on `isPrivate`, etc). Logic below
 * adds delete transcription in "Danger Zone" section

## Indexing logic

| Scenario | Words Cleared | Issues Cleared | Region Re-enqueueing |
|----------|---------------|----------------|---------------------|
| **Transcription Deleted** | ✅ | ✅ | ❌ |
| **`isPrivate`: false → true** | ✅ | ✅ | ❌ |
| **`publicIssues`: true → false** | ❌ | ✅ | ❌ |
| **`publicIssues`: false → true** | ❌ | ❌ | ❌ (handled by `indexRegionData`) |
| **`lang` becomes invalid** | ✅ | ❌ | ❌ |
| **`lang` becomes valid & public** | ❌ | ❌ | ✅ (triggers re-indexing) |
| **`isPrivate`: true → false** | ❌ | ❌ | ✅ (triggers re-indexing) |

## New UI toggle
screenshot-2025-08-14-16 13 50@2x

## Danger zone
screenshot-2025-08-14-19 54 59@2x
screenshot-2025-08-14-19 55 06@2x


amplify/backend/function/indexRegionData/src/lib/dynamo.js
```diff
@@ -17,4 +17,26 @@ async function getDoc(params) {
   })
 }
 
-module.exports = { getDoc }
+async function query(params) {
```
danielfay, line 20: More a note to myself to move to a layer later. 

This file and the one used by `onTranscriptionChange` have a lot of overlap. I suspect there is reuse in other `lib` files as well.

https://github.com/kiyanaw/kiyanaw-backend/issues/171

amplify/backend/function/indexRegionData/src/index.js
```diff
@@ -8,20 +10,161 @@
...
 
 const { okResponse } = require('./utils')
 
+/**
+ * Get all issues for a specific region
+ * @param {string} regionId 
+ * @param {string} issueTable 
+ * @returns {Promise<Array>} Array of issues
+ */
+const getIssuesForRegion = async (regionId, issueTable) => {
```
danielfay: Interesting. So the `search` lib provides functions to do tasks (ex: `search.indexRegionAnalysis`), but the `dynamo` lib just provides different access primitives (ex: `dynamo.scan`)

Do you see any downside to moving this entire function into the `dynamo` lib?

amplify/backend/function/onTranscriptionChange/src/lib/cleanup.js
```diff
@@ -0,0 +1,316 @@
...
+    const params = {
+      TableName: inviteTable,
+      FilterExpression: 'transcriptionId = :transcriptionId',
+      ExpressionAttributeValues: {
+        ':transcriptionId': transcriptionId
+      },
+      ProjectionExpression: 'id'
+    }
+    
+    const result = await dynamo.scan(params)
```
danielfay: From here down in all three if these delete functions the code is identical (other than that the deleted data type is used in the console logs (like, 'invites').

Either `dynamo` lib or this file should provide some kind of general delete.

src/components/auth/AuthenticatorShell.tsx
```diff
@@ -38,10 +38,85 @@ const AuthenticatedApp = ({ children }: { children: ReactNode }) => {
   return <div>{children}</div>;
 };
 
+// Custom theme with just brand colors for highlights
+const kiyanawTheme = createTheme({
```
danielfay, line 43: This strikes me as something that can be initalized somewhere more general and handed to components that use it, but I don't know React well enough to know where to move it.

Ah, the main primary is also here

https://github.com/kiyanaw/kiyanaw-backend/blob/aca9bcae090d36d649f43e01bfacc5567ebeb5c3/src/index.css#L3

Though I'm not sure it's being used.

src/pages/EditorPage.tsx
```diff
@@ -57,9 +58,17 @@ export const EditorPage = () => {
...
+  const handleSaveChanges = (updates: { title?: string; comments?: string; isPrivate?: boolean; publicIssues?: boolean; lang?: string }) => {
     updateTranscription(updates);
   };
+
+  const handleDeleteTranscription = () => {
+    // Add a 2 second pause before navigating
+    setTimeout(() => {
+      // Navigate back to the transcription list after deletion
+      navigate('/transcribe-list/');
+    }, 4000);
```
danielfay, line 135: How strange, this doesn't match the comment is instead 4 seconds.

danielfay (approved):

amplify/backend/function/indexRegionData/src/index.js
aaronfay: Moved to `dynamo.js` and also moved that to use `QUERY` instead of scan (added an index)

src/components/auth/AuthenticatorShell.tsx
aaronfay: I'm gong to make a note of this and come back to it, more work on the theme/layout refactor is needed.

src/pages/EditorPage.tsx
aaronfay: Fixed

amplify/backend/function/onTranscriptionChange/src/lib/cleanup.js
aaronfay: Also using `.scan`, will fix.

aaronfay: Fixed.

PR #166, opened 2025-08-11
Add Issues Support + bonus items

aaronfay:
Main feature here is Issues:

 * create and update issues
 * major interface overhaul to make management intuitive
 * show comment counts inline
 * full issue list on load with buttons to navigate to associated regions
 * syncing text between text editor & issues
 * handling for issues that are out of sync
 * move Issue comments to dedicated model
 * realtime support for issue changes & comment changes

Also quite a few UI cleanup items: 

 * remove unnecessary/unused buttons
 * give more options for where/how to display video
 * condense some parts of the UI for better layout
 * handle smaller screens
 * change isPrivate > is Discoverable

screenshot-2025-08-12-20 33 04@2x
screenshot-2025-08-12-20 33 19@2x


src/components/forms/TranscriptionSettingsPage.tsx
```diff
@@ -86,8 +86,8 @@ export const TranscriptionSettingsPage = ({
       updates.comments = comments;
     }
 
-    if (isPrivate !== (initialIsPrivate ?? true)) {
-      updates.isPrivate = isPrivate;
+    if (isPublic !== (!(initialIsPrivate ?? true))) {
```
danielfay: I don't mind using `isPrivate` in here if it makes the code clearer (and I think it will aha).

I am only really concerned about the label that the user sees being 'Is Discoverable'

src/components/issues/IssueDetailsDialog.tsx
```diff
@@ -0,0 +1,324 @@
...
+      await deleteComment({ commentId });
+    } catch (error) {
+      console.error('Failed to delete comment:', error);
+      // TODO: Show user-friendly error message
+    }
+  };
+
+  // Handle resolving/reopening issue
+  const handleToggleResolved = () => {
+    if (issue && onUpdateIssue) {
```
danielfay, line 106: I was trying to figure out why this and the delete one were optional. I believe it's because if the user can't delete or can't edit then the callback to do so isn't provided. Pretty interesting.

I'm not totally sure the two layers of security are needed though. Looking below if the user can't delete for example, then the markup isn't even rendered to them.

src/components/player/WaveformPlayer.tsx
```diff
@@ -94,6 +97,47 @@ export const WaveformPlayer = ({
     wavesurferService.setPlaybackRate(value);
   };
 
+  // Video control handlers
+  const handleVideoPosition = (position: 'left' | 'center' | 'right') => {
```
danielfay, line 101: This is serviceable for now. I think it is more intuitive to let the user drag the mini-window with the video where ever they want, like if your in a google meeting or on youtube.

src/components/regions/RegionItem.tsx
```diff
@@ -84,7 +92,7 @@ export const RegionItem = ({
...
           </div>
           <div
-            className="pr-4 leading-relaxed [&_.known-word]:text-blue-600 [&_.unknown-word]:bg-red-50 [&_.unknown-word]:text-red-800 [&_.proper-noun]:font-bold [&_.proper-noun]:text-blue-700 [&_.emphasis]:italic [&_.strong]:font-bold empty:before:content-['No_text_content'] empty:before:text-gray-400 empty:before:italic"
+            className="pr-4 leading-relaxed [&_.known-word]:text-blue-600 [&_.unknown-word]:bg-red-50 [&_.unknown-word]:text-red-800 [&_.proper-noun]:font-bold [&_.proper-noun]:text-blue-700 [&_.emphasis]:italic [&_.strong]:font-bold [&_.issue-needs-help]:bg-red-50 [&_.issue-needs-help]:text-red-800 [&_.issue-needs-help]:px-0.5 [&_.issue-needs-help]:rounded [&_.issue-indexing]:bg-yellow-50 [&_.issue-indexing]:text-yellow-800 [&_.issue-indexing]:px-0.5 [&_.issue-indexing]:rounded [&_.issue-new-word]:bg-green-50 [&_.issue-new-word]:text-green-800 [&_.issue-new-word]:px-0.5 [&_.issue-new-word]:rounded [&_.issue-comment-icon]:inline-flex [&_.issue-comment-icon]:items-center [&_.issue-comment-icon]:opacity-70 [&_.issue-comment-icon]:ml-1 empty:before:content-['No_text_content'] empty:before:text-gray-400 empty:before:italic"
```
danielfay, line 95: This `className` is pretty impressive. I haven't worked on the frontend nearly at all so I'm not sure what is recommended, but I'm sure eventually we'll want to start making some reusable CSS classes so that it doesn't need to all in inline like this.

src/hooks/useCreateIssue.ts
```diff
@@ -0,0 +1,11 @@
+import { useCallback } from 'react';
+import { CreateIssueUseCase, type CreateIssueInput } from '../use-cases/create-issue';
+import type { IssueData } from '../services/adt';
+
+const createIssueUseCase = new CreateIssueUseCase();
```
danielfay: Looking at all the hooks in this PR this is the odd one out. All the other ones create the use case within the exported function.

src/hooks/useDeleteIssue.ts
```diff
@@ -0,0 +1,10 @@
+import { useCallback } from 'react';
+import { DeleteIssueUseCase, type DeleteIssueInput } from '../use-cases/delete-issue';
+
+const deleteIssueUseCase = new DeleteIssueUseCase();
```
danielfay: I was mistaken, this hook is also created outside of the function. I don't really see the difference so I were prefer if they all used the same pattern.

src/hooks/useNavigateIssueRegions.ts
```diff
@@ -0,0 +1,17 @@
+import { useCallback } from 'react';
+import { NavigateIssueRegions } from '../use-cases/navigate-issue-regions';
+import type { NavigateDirection } from '../use-cases/navigate-issue-regions';
+import { useEditorStore } from '../stores/useEditorStore';
+
+export const useNavigateIssueRegions = () => {
+  return useCallback((direction: NavigateDirection) => {
+    try {
+      // Get the store - this should work exactly like useSelectAndPlayRegion
```
danielfay: Unneeded comment.

src/hooks/useTextEditors.ts
```diff
@@ -37,17 +39,71 @@ export const useTextEditors = (regionId: string, activeTab: 'main' | 'translatio
...
-      const knownWords = Array.from(useEditorStore.getState().knownWords);
-      if (knownWords.length > 0) {
-        rteService.applyKnownWordsFormatting(mainEditorKey, knownWords);
+    }
+
+    // Always run matching/highlighting on attach to compute link statuses
+    {
+      const state = useEditorStore.getState();
+      const knownWords = Array.from(state.knownWords);
+      if (typeof rteService.updateIssueHighlighting === 'function') {
```
danielfay: Can you explain these `=== 'function'` checks to me please? Is it because they can be null? If so it might be better to check if they exist or not. At least to me that would read less unusual.

Also, what is the logical flow here?

Try to use `updateIssueHighlighting`, but if you can't then try to use `getIssuesForRegion` and `convertIssuesToHighlights`. Then try to use `applyHighlighting`, but if you can't then fall back to using `applyKnownWordsFormatting`?

src/hooks/useTextEditors.ts
```diff
@@ -67,10 +123,30 @@ export const useTextEditors = (regionId: string, activeTab: 'main' | 'translatio
...
-        // IMMEDIATELY apply cached known words formatting
+        // IMMEDIATELY apply highlighting and run issue matching detection
         // This solves format inheritance and word splitting issues
-        const knownWords = Array.from(useEditorStore.getState().knownWords);
-        rteService.applyKnownWordsFormatting(mainEditorKey, knownWords);
+        const state = useEditorStore.getState();
+        const knownWords = Array.from(state.knownWords);
+        
+        // Use updateIssueHighlighting which includes matching service detection
+        if (typeof rteService.updateIssueHighlighting === 'function') {
```
danielfay: I believe this is duplicate code of the block I commented on above at R48 above.

src/hooks/useUpdateIssue.ts
```diff
@@ -0,0 +1,11 @@
+import { useCallback } from 'react';
+import { UpdateIssueUseCase, type UpdateIssueInput } from '../use-cases/update-issue';
+import type { IssueData } from '../services/adt';
+
+const updateIssueUseCase = new UpdateIssueUseCase();
```
danielfay: A third one declared outside the exported function.

src/components/issues/IssueDetailsDialog.tsx
```diff
@@ -0,0 +1,324 @@
...
+
+  // Don't render if not open or issue not found
+  if (!isOpen || !issue) {
+    return null;
+  }
+
+  const typeInfo = getIssueTypeInfo(issue.type);
+  const isIssueOwner = user?.userId === issue.owner;
+  const userCanComment = canComment(transcription, user);
+  const userCanEdit = transcription && user ?
```
danielfay: I'm not as sure about all the other permission Booleans here, but it feels like this one can definitely move into the permission lib.

src/pages/EditorPage.tsx
```diff
@@ -48,8 +60,30 @@ export const EditorPage = () => {
     updateTranscription(updates);
   };
   
-  // Determine if current user is the owner
+  // Determine user permissions for this transcription
   const isOwner = transcription?.author === user?.userId;
+  const isEditor = transcription?.editors?.includes(user?.userId || '') || false;
```
danielfay: I'm sure there is stuff like this all over. As a more general question. What is your endgame goal with the permission lib?

Personally I would like to see all of these end up in there eventually.

It looks like these two are done to create `canEdit`, which to me makes a strong case for `canEditTranscription(transcription, user)` to be provided by the permission lib.

src/services/adt.ts
```diff
@@ -326,6 +340,117 @@ export class RegionModel {
...
+   */
+  isMine(currentUserId?: string): boolean {
+    if (!currentUserId) return false;
+    return this.author === currentUserId;
+  }
+
+  /**
+   * Get the display name for the author, showing "me" if it's the current user
+   */
+  getAuthorDisplay(currentUserId?: string): string {
```
danielfay, line 410: Nice! I like this. I noticed this being done in one of the files above manually in the markup. I'll use this as an example going forward and switch them if I see them.

src/services/issueHighlightService.ts
```diff
@@ -0,0 +1,53 @@
...
+   * Convert issue data to highlight format for text highlighting
+   */
+  convertIssuesToHighlights(issues: IssueData[]): IssueHighlight[];
+}
+
+class IssueHighlightServiceImpl implements IssueHighlightService {
+  /**
+   * Map issue type string to IssueType enum
+   */
+  private mapIssueType(typeString: string): IssueType {
```
danielfay: Is this still needed? I thought I saw you selecting the issue type by clicking a little badge.

src/services/issueMatchingService.ts
danielfay, line 1: This service is quite sophisticated. 

Is it doing something similar to the 'suggest' endpoint of `spellcheck`?

src/services/issueService.ts
```diff
@@ -2,53 +2,249 @@ import { generateClient } from 'aws-amplify/api';
...
-    // Process issues: parse comments and sort
-    return issues.map((issue: IssueData): ProcessedIssue => {
-      let comments: IssueComment[] = [];
-      try {
-        comments = issue.comments ? JSON.parse(issue.comments) : [];
-      } catch {
-        console.warn('Failed to parse comments for issue:', issue.id);
-        comments = [];
+    // Sort issues by creation date (newest first)
+    return issues.sort(
```
danielfay, line 51: I could've sworn that I saw this sort being done again on the page that uses it. This is actually my preferred way of doing it. Returning the 'ideal' sort so that the front end doesn't need to worry about it.

src/services/textHighlightService.ts
```diff
@@ -4,13 +4,33 @@ export interface HighlightMatch {
   length: number;
 }
 
+export type IssueType = 'needs-help' | 'indexing' | 'new-word';
```
danielfay, line 9: I think these are in a few places. They can probably move to `types/shared.ts` since a few files need them.

src/services/textHighlightService.ts
```diff
@@ -24,17 +44,59 @@ class TextHighlightServiceImpl implements TextHighlightService {
...
+    const { knownWords = new Set(), issues = [] } = options;
+    
+    if (knownWords.size === 0 && issues.length === 0) {
+      return text;
     }
 
+    // Create issue text lookup with type and comment information for efficient matching
+    // Normalize to extract letters/numbers/_/- only (consistent with tokenization)
+    const issueTextMap = new Map<string, { type: IssueType; commentCount: number }>();
+    const tokenPattern = /([\p{L}\p{N}_-]+)/u;
```
danielfay: This token pattern is in 5 or 6 different places in this PR alone. I'm getting the impression that everything needs to break things apart into tokens the same way, so this should probably be put somewhere that they can all import.

src/services/textHighlightService.ts
```diff
@@ -24,17 +44,59 @@ class TextHighlightServiceImpl implements TextHighlightService {
...
+    issues.forEach(issue => {
```
danielfay, line 65: Maybe the `issueMatchingService` can provide this functionality? I'm surprised a `textHighlightService` has such intimate knowledge about how to tokenize issues.

Looking back at the `rteService` it also does something very similar to issues. Maybe the `issueService` itself should provide tokens to each of them?

src/services/userService.ts
```diff
@@ -4,6 +4,14 @@ export const currentUser = (): AuthUser | null => {
   return useAuthStore.getState().user;
 };
 
+export const currentUserFriendly = (): string | null => {
```
danielfay, line 7: Awesome! I think there are a handful of places (even some in this PR) that are doing this on there own. I'll keep this in mind and replace them as I see them and touch code 😄

src/use-cases/analyze-region-text.ts
```diff
@@ -117,10 +118,25 @@ export class AnalyzeRegionTextUseCase {
     // Update region analysis in store (this will be picked up by the coordinated save)
     store.setRegionAnalysis(regionId, allKnownWords);
 
-    // Apply formatting to the main editor
+    // Apply known words and issue highlighting to the main editor
     const mainEditorKey: EditorKey = `${regionId}:main`;
-    if (rteService.hasEditor(mainEditorKey)) {
-      rteService.applyKnownWordsFormatting(mainEditorKey, allKnownWords);
+    const configRteService = this.config.services.rteService;
+    if (configRteService.hasEditor(mainEditorKey)) {
```
danielfay, line 124: This pattern of `=== function` ifs looks familiar. Maybe one of the highlight services needs to provide a short cut in doing this?

src/use-cases/create-issue.ts
```diff
@@ -0,0 +1,86 @@
...
+        text: input.text.trim(),
+        type: input.type,
+        owner: input.owner,
+        ownerFriendly: input.ownerFriendly,
+        regionId: input.regionId || '',
+        transcriptionId: input.transcriptionId,
+      });
+
+      // Replace optimistic issue with real one from backend
+      store.deleteIssue(optimisticIssue.id);
```
danielfay, line 83: Interesting, I'm surprised this doesn't make the interface flicker. Could you also not just do an update with the returned issue from the backend?

src/use-cases/delete-issue.ts
```diff
@@ -0,0 +1,47 @@
...
+    }
+  }
+
+  async execute(input: DeleteIssueInput): Promise<void> {
+    this.validate(input);
+
+    try {
+      // Get the issue and its version before deleting it
+      const store = useEditorStore.getState();
+      const issue = store.issues.find(i => i.id === input.issueId);
```
danielfay: `commentById` exists. Any reason not to have an `issueById`?

src/use-cases/load-transcription.ts
```diff
@@ -65,13 +70,26 @@ export class LoadTranscription {
...
+    } else if (selectedIssueId) {
+      // If only issueId is present, seek to its region if found
+      const issue = data.issues?.find((i) => i.id === selectedIssueId);
+      const regionIdFromIssue = issue?.regionId;
+      if (regionIdFromIssue) {
+        const selectedRegion = data.regions.find((r) => r.id === regionIdFromIssue);
+        if (selectedRegion) {
+          wavesurferService.seekToRegion(selectedRegion);
+          browserService.setSelectedRegion(regionIdFromIssue);
+          if (this.config.store.setSelectedRegion) {
```
danielfay, line 88: It should be safe to drop this `if`

src/use-cases/load-transcription.ts
```diff
@@ -41,11 +41,16 @@ export class LoadTranscription {
...
     
-    // Check if there's a regionId in the URL that we should select
+    // Read deep-link params
     const selectedRegionId = browserService.getRegionIdFromUrl();
+    const selectedIssueId = browserService.getIssueIdFromUrl?.() || null;
     
     // Set transcription data in store - use null instead of undefined for consistency with tests
     this.config.store.setFullTranscriptionData(data, selectedRegionId || null);
+    // Persist selected issue id in store if provided
+    if (selectedIssueId && this.config.store.setSelectedIssueId) {
```
danielfay, line 51: Is there a case where the store isn't provided? I think this is another case of Claude being too defensive. You still need to check for `selectedIssueId` here though since it can be `null`.

src/use-cases/subscribe-to-comment-changes.ts
```diff
@@ -0,0 +1,157 @@
+import { services } from '../services';
+import type { CommentSubscriptionEvent } from '../services/commentService';
+import { issueHighlightService } from '../services/issueHighlightService';
+import Timeout from 'smart-timeout';
+
+export interface SubscribeToCommentChangesConfig {
+  transcriptionId: string;
+  services: typeof services;
```
danielfay, line 8: Seeing this reminded me. Maybe `stores/*` should also have an `index.ts` so that you don't need to type it as `any`.

.gitignore
```diff
@@ -50,3 +50,4 @@ amplifytools.xcconfig
 #amplify-do-not-edit-end
 
 coverage/
+scripts/migrate-data
```
danielfay, line 53: We had talked about not wanting to expose data from the notebooks in a recent call. Maybe we should put project utilities in a private Cloudcraft repo or something so that we both have access.

It looks like something like https://github.com/srstevenson/nb-clean could be useful as well.

danielfay (approved):
Nice work and getting issues and comments working!

Half of these comments are minor. Some others are more general questions about how certain services work and what they should be responsible for going forward. Nothing blocking, so work through the feedback as you see fit.

Overall I'm getting better at navigating and reading the code base which feels nice.

src/components/forms/TranscriptionSettingsPage.tsx
aaronfay: good call.

src/components/player/WaveformPlayer.tsx
aaronfay: Good idea. I wonder what it would take to do "pop out window" too, that would be cool.

src/components/issues/IssueDetailsDialog.tsx
aaronfay: Hmm, not sure how I feel about this one, I'm going to leave this piece for now.

src/hooks/useCreateIssue.ts
aaronfay: Yeah good catch

src/hooks/useDeleteIssue.ts
aaronfay: Yup good catch.

src/hooks/useTextEditors.ts
aaronfay: No, this block makes no sense, it should be a one-liner. I think Claude got caught in a loop trying to fight a bug or something and hallucinated a single line of code into the most ridiculous thing possible. I see a bunch of these in the codebase, cleaning up now.

aaronfay: I've cleaned all of these "defensive" checks up.

src/components/regions/RegionItem.tsx
aaronfay: I _think_ we've come "full circle" to inline classes like this. Could be wrong but I think the whole "make named classes for everything" has been deemed unsustainable and the preference nowadays (unless you're building a design system) is utility classes. The industry might be split a little bit but I _believe_ This Is The Way™

src/hooks/useTextEditors.ts
aaronfay: This has been cleaned up.

src/hooks/useUpdateIssue.ts
aaronfay: Fixed

src/components/issues/IssueDetailsDialog.tsx
aaronfay: Fixed

src/pages/EditorPage.tsx
aaronfay: fixed

src/services/issueHighlightService.ts
aaronfay: removed.

src/services/issueMatchingService.ts
aaronfay: Not quite, the entire service basically determines how to match issue text to region data, so we have an "issue" with  the text "problemword" and then the region text "this region has a problemword" so the service needs to work out what needs to match. 

But from time to time (because of legacy system jank/data) they don't completely match, so the sophisticated part of the service is so that when the issue text _doesn't_ match anything in the region, we can make a suggestion to help clean it up. 

We show a little ⚠️ icon and if they click it they get a little dropdown to suggest how to link it back up to text.

screenshot-2025-08-13-13 49 31@2x

src/services/textHighlightService.ts
aaronfay: Will make a ticket.

src/services/userService.ts
aaronfay: Thank you!

src/use-cases/analyze-region-text.ts
aaronfay: I think this one is legit, because only one editor can be present at a time (currently). I'm inclined to leave this one for now.

src/use-cases/create-issue.ts
aaronfay: Hmm, it _was_ saving first and then popping the issue in, I asked Claude to show the issue right away, because we don't display the ID anywhere I think it's fine 🤷🏻

src/services/textHighlightService.ts
aaronfay: Fixed

src/services/textHighlightService.ts
aaronfay: No, the issues _are_ highlighted, different bg colors for different issue types. Because we highlight in both the RegionList as well as the RTE that is why this service came into existence, because we had disparate code with separate matching logic. And the issue is (from experience) we will run into more and more edge cases around matching, so we need a common place with tests to make sure that we can handle them.

src/use-cases/load-transcription.ts
aaronfay: I think this is outdated, I have more recent changes.

src/use-cases/subscribe-to-comment-changes.ts
aaronfay: Goona move this to a ticket.

.gitignore
aaronfay: Yup, I think that's a good idea.

PR #167, opened 2025-08-12
Successful fresh deploy to production

danielfay:
The most important learning from this work is that if you add new backend functionality that has dependancies on other amplify resources you need to run `npx amplify env checkout <env>` to regenerate the file that `amplify push` uses. I just realized this more when I stashed some other work that isn't working that if you have phantom infra even though the files have been stashed that command will also fix it (likely for the same reason ie. that `amplify status` looks at the generated file and not the current state of the amplify folder)

Very few changes were needed to get production deploying. Mostly just properly referencing the open search dependency for the index function. The region SQS also properly waits for it's function to be created now.

All other changes were made by amplify. I'm hoping that the `env` name changes to the lambda layers are benign. We'll have to see the next time we deploy backend changes to staging.

### DIFF
--- a/amplify/backend/api/APIGatewayAuthStack.json
+++ b/amplify/backend/api/APIGatewayAuthStack.json
@@ -11,10 +11,10 @@
     "env": {
       "Type": "String"
     },
-    "spellcheck": {
+    "invite": {
       "Type": "String"
     },
-    "invite": {
+    "spellcheck": {
       "Type": "String"
     }
   },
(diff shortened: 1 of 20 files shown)


aaronfay:
I _think_ they way we're going to want to do this is, instead of `amplify env checkout staging` or whatever, then deploy and commit changes, then `amplify env checkout production` and commit those changes, I'm inclined to say let's _not_ commit the production changes to `git`. They will technically exist when we `checkout`/`pull` that environment, but what if we just keep the staging changes in `git` so that those files aren't just a bunch of commits thrashing back and forth, does that seem reasonable?

danielfay:
Good idea. That actually behaved quite well. These are the changes after a `npx amplify env checkout staging`

Amplify just reordered some stuff, but the layers are no longer thrashing which was also worrying me.

aaronfay (approved):

PR #169, opened 2025-08-14
Production fixes

aaronfay:
A small smattering of "production" fixes:

 - not all regions loading correctly (using wrong query)
 - transcription list doesn't scroll
 - settings page is not scrollable
 - invites page is not scrollable
 - modifies Transcription list from table to card view
 - add "Public" to transcription list
 - vertical videos don't look great
 - ability to minimize video

The biggest change is moving the "table" to a list of cards for the Transcription list, which helps support more viewport options:
screenshot-2025-08-14-09 42 02@2x


danielfay (approved):
Sweet, these changes look good.

PR #173, opened 2025-08-16
Add counts to transcription list

aaronfay:
Adds `regionCount` and `issueCount` to the transcription list, as well as updates a bunch of the use-cases to make sure that those counts are updated appropriately when different things change (eg: add/remove issue)

screenshot-2025-08-16-13 37 56@2x


danielfay (approved):
Looks great!

PR #175, opened 2025-08-17
Adds PWsupport

aaronfay:
I was playing around on mobile and thought, "wonder what it would take to enable PWA support?" turns out not very much.

PWA makes the website "installable" on mobile devices and provides access to a few more features (like push notifications, etc). "Installable" is mainly what we're doing here, removes the browser nav bar and just provides a nicer experience for the user. Down the road we'll be able to provide nicer offline support, etc.

Tweaks how regions are created on mobile. 
Also disables "create" on mobile for now until we can support more audio/video formats.

image

image


src/components/player/WaveformPlayer.tsx
```diff
@@ -149,22 +156,86 @@ export const WaveformPlayer = ({
...
+    } else {
+      // Activate sparkle mode
+      setIsSparkleActive(true);
+    }
   };
 
+  // Create a new region using the mobile sparkle functionality
+  const createMobileRegion = useCallback(async (start: number, end: number) => {
+    try {
+      const regionId = `region-${Date.now()}-${Math.random().toString(36).substr(2, 9)}`;
```
danielfay: I found it a little strange that the WaveformPlayer knew how to generate a `regionId`, but after looking around for how we generate the id when a new region is created I realized we are using the id wavesurfers provides.

Which is probably this: 

https://github.com/katspaugh/wavesurfer.js/blob/d2eaebbea4ccf1933a96bd7cdad572d27577c214/src/plugins/regions.ts#L113

A shame there isn't a wavesurfer utility for generating an id.

danielfay: Actually, no, I'm still confused.

The ids in the dynamodb table look like this `wavesurfer_ru7m02ccgl8`

There are also two recent regions in the staging table similar to this `region-5ba1pcggc0k`, though if you were playing with making regions on your phone I would expect them to be different (and have the date like above)

danielfay (approved):
Looks cool, the code itself all seems pretty straight forward.

I am however left mystified by how regions (and specifically region ids) work in this project.

src/components/player/WaveformPlayer.tsx
danielfay: Interesting. My new regions from local dev to staging are also in the `region-*` format.

```
Region Created 
Object { id: "region-82unpg8d3ls", start: 14.53789409107402, end: 20.702297808417995 }
useWavesurferEvents.ts:58:15
```

Funnily this is what I would assume it always was based on the wavesurfer code. Maybe the `wavesurfer_*` ids are from an old very of the library or something.

danielfay (approved):

src/components/player/WaveformPlayer.tsx
aaronfay: So yeah actually we shouldn't be generating region IDs at all, coz wavesurfer gives them to us for free. The _old_ version of wavesurfer used `wavesurfer-` and the _new_ version of wavesurfer seems to use `region-` so  🤷🏻 I just use whatever it gives us. 

Not sure how we ended up generating IDs here, we shouldn't need it. I'll take another look.

danielfay: Sweet, thanks.

PR #177, opened 2025-08-20
Fixes the missing "Add new +" transcription button on desktop

aaronfay:
Shut off on mobile for now, accidentally shut off desktop. This was a hotfix, and is in production already.


danielfay (approved):
Looks good. Sorry for missing this in the review, I think I need to start reviewing in my ID again so that I can better scan the context of the whole file/section.

As an aside, there is a second use of `showUploadButton` in that file. Does it need to be adjusted as well?

PR #183, opened 2025-08-30
Handle huge peaks files by chunking

aaronfay:
The old naiive code would just crap out with a massive (say 500kb) json file, this handles it in chunks for large peaks files.

### DIFF
--- a/amplify/backend/function/createPeaksFile/src/lib/audio.js
+++ b/amplify/backend/function/createPeaksFile/src/lib/audio.js
@@ -73,7 +73,7 @@ const processPeaksData = async (jsonPath) => {
   console.log(`Max peak value: ${max}`)
   
   // Normalize data using chunked processing to avoid stack overflow
-  parsed.data = processLargeArray(parsed.data, max)
+  parsed.data = processArrayInChunks(parsed.data, max)
   
   return parsed
 }
@@ -99,7 +99,7 @@ const getMaxValue = (arr) => {
  * @param {number} max - Maximum value for normalization
  * @returns {number[]} - Processed array
  */
-const processLargeArray = (data, max) => {
+const processArrayInChunks = (data, max) => {
   const chunkSize = 10000
   const result = []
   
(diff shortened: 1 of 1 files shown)


aaronfay:
Tested with a 1-hour long transcription on staging.

amplify/backend/function/createPeaksFile/src/lib/audio.js
```diff
@@ -66,18 +66,61 @@ const processPeaksData = async (jsonPath) => {
...
+  // Find maximum value for normalization (avoid spread operator for large arrays)
+  const max = getMaxValue(parsed.data)
   console.log(`Max peak value: ${max}`)
   
-  // Normalize data to 0-1 range with 2 decimal places
-  parsed.data = parsed.data.map(value => {
-    return Number(Number(value / max).toFixed(2))
-  })
+  // Normalize data using chunked processing to avoid stack overflow
+  parsed.data = processLargeArray(parsed.data, max)
```
danielfay: I think just `processArray` fits better here (there doesn't seem to be a method for processing 'small' or 'normal' arrays).

Then something like 'chunking is used to support large arrays can be added to the doc string below.

danielfay (approved):

amplify/backend/function/createPeaksFile/src/lib/audio.js
aaronfay: My reasoning is, "what is 'normal'"? I don't think there's harm in using the same operation just by default, one less logic path if computationally it doesn't make much difference.

aaronfay: Oh I see, yes. There _was_ a normal function haha. Let me clean that up

danielfay: Aha, glad you realized what I was saying. Sorry if it was a little unclear.

I'm all for there being one robust path through the code. It was just named like that _wasn't_ the case, which I found confusing :)

PR #190, opened 2025-09-06
Fixes full SCAN of transcriptions on load

aaronfay:
This PR modifies the way that we load transcriptions for a user. 

By using the current `listTranscriptions` we incur a full `SCAN` of the DynamoDB table for `Transcription`, which creates an issue where if a user has 2 Transcriptions at the top of the table but one Transcription at the bottom of the table (more than 100 rows later) then we need to re-query with `nextToken` to get the full list. This scans the whole DynamoDB table for every listing, and is both inefficient and will become costly. Currently in production we don't do this at all, so some Transcriptions for some users are just "missing".

To mitigate this, we're changing the way we pull the data

 * query to get "My transcriptions" by using a new index on the `author` field,
 * query to get "Shared with me" by querying Invites. 

In the UI, similarly we'll split these into two tabs. 

screenshot-2025-09-06-11 31 46@2x

However, in order for Invites to show "Transcription"-like data, we've modified the `handle-get-my-invites` lambda to 

 * a) pull all the invites for the user, and
 * b) do a bulk query against the corresponding Transcription IDs to get supplemental information so the response data can be used to show the "Shared with Me" tab. 

Lastly, the PR introduces some "local caching" to the app, so on first load the Transcriptions and Invites are retrieved in full and stored locally in an indexed db (using Dexie). A "last sync timestamp" is created and every subsequent load queries against the GSI for both Transcription and Invite to _only_ pull any data that was created (for Invites) and updated (for Transcriptions) since that timestamp. There is a little complexity here because Invites are managed through the Lambda and not through the normal API methods.

The result is that users see data _instantly_ and the "sync" is done in the background, resulting in all user Transcriptions showing up, but also a much nicer UX. Eventually we will also use local caching for Region and Peaks data, but we'll leave that for another PR.

### DIFF
--- a/amplify/backend/function/inviteHandler/src/actions/services/transcription-service.js
+++ b/amplify/backend/function/inviteHandler/src/actions/services/transcription-service.js
@@ -65,7 +65,6 @@ class TranscriptionService {
 
         totalRetrieved += transcriptions.length;
         
-        // Handle any unprocessed keys (due to throttling)
         if (result.UnprocessedKeys && Object.keys(result.UnprocessedKeys).length > 0) {
           console.warn(`⚠️ Some keys were unprocessed in batch get, may need retry logic`);
         }
--- a/amplify/backend/function/inviteHandler/src/actions/services/transcription-service.js
+++ b/amplify/backend/function/inviteHandler/src/actions/services/transcription-service.js
@@ -1,0 +1,114 @@
+const { DynamoDBClient } = require('@aws-sdk/client-dynamodb');
+const { DynamoDBDocumentClient, BatchGetCommand } = require('@aws-sdk/lib-dynamodb');
+const { InternalError } = require('../errors/invite-errors');
+
+/**
+ * Transcription Service for Lambda
+ * 
+ * Handles batch retrieval of transcription data for invite processing
+ */
+class TranscriptionService {
+  constructor() {
+    // Initialize DynamoDB client
+    const client = new DynamoDBClient({ region: process.env.REGION });
+    this.docClient = DynamoDBDocumentClient.from(client);
+    this.tableName = process.env.API_KIYANAW_TRANSCRIPTIONTABLE_NAME;
+  }
+
+  /**
+   * Batch get transcriptions by IDs with efficient chunking
+   * @param {string[]} transcriptionIds - Array of transcription IDs to retrieve
+   * @returns {Object} Map of transcriptionId -> transcription data
+   */
+  async getTranscriptionsByIds(transcriptionIds) {
+    if (!this.tableName) {
+      throw new InternalError('API_KIYANAW_TRANSCRIPTIONTABLE_NAME environment variable not configured');
+    }
+
+    if (!transcriptionIds || transcriptionIds.length === 0) {
+      console.log('📊 No transcription IDs provided, returning empty map');
+      return {};
+    }
+
+    console.log(`📊 Batch loading ${transcriptionIds.length} transcriptions...`);
+
+    // DynamoDB BatchGetItem has a limit of 100 items per request
+    const chunks = this.chunkArray(transcriptionIds, 100);
+    const allTranscriptions = {};
+    let totalRetrieved = 0;
+
+    for (const chunk of chunks) {
+      const keys = chunk.map(id => ({ id }));
+      
+      const command = new BatchGetCommand({
+        RequestItems: {
+          [this.tableName]: {
+            Keys: keys,
+            // Only get the fields we need for card rendering to reduce data transfer
+            ProjectionExpression: 'id, title, author, authorFriendly, #type, #length, coverage, issueCount, regionCount, commentCount, isPrivate, dateLastUpdated, userLastUpdated, createdAt, updatedAt',
+            ExpressionAttributeNames: {
+              '#type': 'type',
+              '#length': 'length'
+            }
+          }
+        }
+      });
+
+      try {
+        const result = await this.docClient.send(command);
+        const transcriptions = result.Responses[this.tableName] || [];
+        
+        // Build lookup map
+        transcriptions.forEach(transcription => {
+          allTranscriptions[transcription.id] = transcription;
+        });
+
+        totalRetrieved += transcriptions.length;
+        
+        // Handle any unprocessed keys (due to throttling)
+        if (result.UnprocessedKeys && Object.keys(result.UnprocessedKeys).length > 0) {
+          console.warn(`⚠️ Some keys were unprocessed in batch get, may need retry logic`);
+        }
+
+      } catch (error) {
+        console.error(`❌ Error in batch get for chunk:`, error);
+        throw new InternalError(`Failed to retrieve transcription data: ${error.message}`);
+      }
+    }
+
+    console.log(`✅ Batch retrieved ${totalRetrieved} transcriptions out of ${transcriptionIds.length} requested`);
+    
+    // Log any missing transcriptions for debugging
+    const missingIds = transcriptionIds.filter(id => !allTranscriptions[id]);
+    if (missingIds.length > 0) {
+      console.warn(`⚠️ Missing transcriptions for IDs: ${missingIds.join(', ')}`);
+    }
+
+    return allTranscriptions;
+  }
+
+  /**
+   * Utility to chunk array into smaller arrays for batch processing
+   * @param {Array} array - Array to chunk
+   * @param {number} size - Size of each chunk
+   * @returns {Array[]} Array of chunks
+   */
+  chunkArray(array, size) {
+    const chunks = [];
+    for (let i = 0; i < array.length; i += size) {
+      chunks.push(array.slice(i, i + size));
+    }
+    return chunks;
+  }
+
+  /**
+   * Get the configured table name for debugging
+   * @returns {string} The DynamoDB table name
+   */
+  getTableName() {
+    return this.tableName;
+  }
+}
+
+// Export singleton instance
+module.exports = new TranscriptionService();
--- a/src/hooks/useLoadTranscriptions.ts
+++ b/src/hooks/useLoadTranscriptions.ts
@@ -28,7 +28,6 @@ export const useLoadTranscriptions = () => {
       // Fire and forget - let use-case handle the orchestration
       useCase.execute().catch((error) => {
         console.error('❌ useLoadTranscriptions failed:', error);
-        // Error is already handled in the use-case
       });
     }
   }, []);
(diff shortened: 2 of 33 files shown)


amplify/backend/function/inviteHandler/src/actions/services/transcription-service.js
```diff
@@ -0,0 +1,114 @@
...
+        const transcriptions = result.Responses[this.tableName] || [];
+        
+        // Build lookup map
+        transcriptions.forEach(transcription => {
+          allTranscriptions[transcription.id] = transcription;
+        });
+
+        totalRetrieved += transcriptions.length;
+        
+        // Handle any unprocessed keys (due to throttling)
```
danielfay: This looks like a TODO to me, not a comment

src/hooks/useLoadTranscriptions.ts
```diff
@@ -1,25 +1,55 @@
...
+        services,
+        store,
+      });
 
-    useCase.execute()
-  }
+      // Fire and forget - let use-case handle the orchestration
+      useCase.execute().catch((error) => {
+        console.error('❌ useLoadTranscriptions failed:', error);
+        // Error is already handled in the use-case
```
danielfay: I think I asked this question before in another PR. If the use-case handles the error then why is it caught here?

src/services/transcriptionService.ts
```diff
@@ -2,7 +2,66 @@ import { generateClient } from 'aws-amplify/api';
 import { getUrl } from 'aws-amplify/storage';
 // eslint-disable-next-line @typescript-eslint/ban-ts-comment
 // @ts-ignore - GraphQL queries are generated as JS files
-import { getTranscription, listTranscriptions } from '../graphql/queries.js';
+import { getTranscription, transcriptionsByAuthor } from '../graphql/queries.js';
+
+// Custom query for compound GSI (author + dateLastUpdated)
+const transcriptionsByAuthorDate = /* GraphQL */ `
```
danielfay: Please move this below the imports. Or maybe better yet, make a `customQueries.js` in the `graphql` folder and move it there.

src/services/transcriptionService.ts
```diff
@@ -255,37 +316,560 @@ export const loadInFull = async (transcriptionId: string): Promise<false | LoadT
...
+            nextToken?: string;
+          }
+        } 
+      };
+      const page = response.data?.transcriptionsByAuthor;
+      
+      if (page?.items) {
+        const user = currentUser();
+        const pageModels = page.items.map(item => {
+          const model = new TranscriptionModel(item as unknown as ADTTranscriptionData);
```
danielfay, line 299: I'm really surpised that an `as unknown as` work around was needed here. `SharedTranscriptionData` looks like it is literally just an alias for `ADTTranscriptionData` from the `shared.ts`

src/services/transcriptionService.ts
```diff
@@ -255,37 +316,560 @@ export const loadInFull = async (transcriptionId: string): Promise<false | LoadT
...
+    // Convert to TranscriptionModel instances using the joined data from Lambda
+    const sharedTranscriptions: TranscriptionModel[] = [];
+    const user = currentUser();
+    
+    for (const inviteWithValidation of acceptedInvitesWithData) {
+      try {
+        const { transcription } = inviteWithValidation;
+        
+        // Type guard to ensure transcription exists
+        if (!transcription) {
```
danielfay, line 378: I suppose I don't mind this eing here, but this is already garenteed when creating `acceptedInvitesWithData` with the filter above

src/services/transcriptionService.ts
```diff
@@ -255,37 +316,560 @@ export const loadInFull = async (transcriptionId: string): Promise<false | LoadT
...
+          locations: gqlError.locations
+        });
+      });
+    }
+    throw new Error(`Failed to load owned transcriptions since ${sinceDate}: ${error}`);
+  }
+};
+
+// Sync state management
+let currentSyncOperation: Promise<TranscriptionModel[]> | null = null;
```
danielfay: Is this a global var for this file? Can you move it near the top then please.

`syncLatestChanges` checking to see if it existed, but not being passed it really threw me for a loop

src/services/transcriptionService.ts
```diff
@@ -255,37 +316,560 @@ export const loadInFull = async (transcriptionId: string): Promise<false | LoadT
...
+  } catch (error) {
+    console.warn('⚠️ Cache validation failed, continuing with existing cache:', error);
+  }
+};
+
+/**
+ * Performs a hybrid full sync using efficient GSI queries + invite discovery
+ * @param sinceTimestamp Optional timestamp for incremental sync
+ */
+const performFullSync = async (sinceTimestamp?: string): Promise<TranscriptionModel[]> => {
```
danielfay: I'm not very fond of this name. It is not infact a full sync when `sinceTimestamp` is provided.

The tension in the name can be seen in the code itself as well. Load transcriptions has two versions `loadOwnedTranscriptionsSince` and `loadOwnedTranscriptions` that is flow controlled with an `if` while `loadSharedTranscriptions` is similar to this method and can take an optional `sinceTimestamp`.

All that being said, the only change I would like is for the method to be `performSync` since I see that it already sets a `syncType` just like `loadSharedTranscriptions` does.

src/services/transcriptionService.ts
```diff
@@ -255,37 +316,560 @@ export const loadInFull = async (transcriptionId: string): Promise<false | LoadT
...
 
-    console.log(`✅ Loaded ${transcriptionModels.length} transcriptions`);
-    return transcriptionModels;
+  const syncType = sinceTimestamp ? 'incremental' : 'full';
+  console.log(`🔍 Performing hybrid ${syncType} sync...`);
+  
+  try {
+    // Load owned transcriptions via efficient GSI
+    let ownedTranscriptions: TranscriptionModel[] = [];
+    if (sinceTimestamp) {
```
danielfay: I think the code would be a bit more clear of this if was based on `syncType`. Then you also don't need the `syncType` style comments above each of the calls below

src/services/transcriptionService.ts
```diff
@@ -255,37 +316,560 @@ export const loadInFull = async (transcriptionId: string): Promise<false | LoadT
...
+      ownedTranscriptions = await loadOwnedTranscriptionsSince(user.userId, sinceTimestamp);
+    } else {
+      // Full sync for owned transcriptions
+      ownedTranscriptions = await loadOwnedTranscriptions(user.userId);
+    }
+    console.log(`📝 Loaded ${ownedTranscriptions.length} owned transcriptions`);
+    
+    // Load shared transcriptions via invites (need email for invite lookup)
+    let sharedTranscriptions: TranscriptionModel[] = [];
+    if (user.username) { // username is typically the email in Cognito
```
danielfay: I don't think it is possible for us not to have this. Mostly since we have email based sign up

src/services/transcriptionService.ts
```diff
@@ -320,6 +904,23 @@ export const create = async (data: CreateTranscriptionData): Promise<SharedTrans
       throw new Error('Failed to create transcription - no data returned');
     }
 
+    // Invalidate cache after successful creation
```
danielfay, line 845: It looks like `create` and `updateTranscription` use the exact same code to update and invalidate the cache. Should probably remove the duplication incase it is needed again

src/use-cases/load-transcriptions-from-cache.ts
danielfay, line 1: All together this works out quite clean. Well done!

src/use-cases/sync-latest-transcriptions.ts
```diff
@@ -0,0 +1,81 @@
...
+  }
+
+  async execute(): Promise<TranscriptionModel[]> {
+    this.validate();
+
+    const transcriptionService = this.config.services.transcriptionService;
+    const { sinceTimestamp, isFullSync } = this.config;
+    
+    try {
+      console.log(`🔄 SyncLatestTranscriptions use-case executing (${isFullSync ? 'full' : 'incremental'} sync)...`);
```
danielfay, line 37: I'm a little surpsed that this is needed just for a console log. I kinda understand why it's behaving different than the other `full` and `incremental` checks, but if checking if `sinceTimestamp` exists is enough in the service shouldn't it be enough here to decide on the `syncType`?

Funnily `syncLatestChanges` doesn't seem to care one way or the other, so this is just for this console log.

edit: Reading the other two `sync-*` use-cases they seem to grab their own cache stats for the timestamp (and sync type). If we do that in this use-case as well then we can simplify how this is called in `load-transcriptions` and remove those two params from the `config`

danielfay (changes requested):
Overall this looks awesome, the only main thing I would like done is a function name change.

src/hooks/useLoadTranscriptions.ts
aaronfay: I've removed the silly comment. There is a case where the use-case throws, so not having "uncaught promise exception" is desirable here.

src/services/transcriptionService.ts
aaronfay: Good call.

src/services/transcriptionService.ts
aaronfay: It should be, this code path should never be hit for sure. Claude being defensive 🤷🏻 I don't think it hurts anything.

src/services/transcriptionService.ts
aaronfay: Yup good call

src/services/transcriptionService.ts
aaronfay: Good call.

src/services/transcriptionService.ts
aaronfay: In this case I'm ok with the duplication. There's not a ton of "logic" in there or anything complex, for me adding another method feels like "adding complexity" or making it less readable. If we get a 3rd time then sure, but for now I feel like this clearer, even if it's duplicated.

src/services/transcriptionService.ts
aaronfay: Yup, symptom of Claude only looking at "part" of a file I think. Fixed.

src/services/transcriptionService.ts
aaronfay: Renamed to `performSync`. Good call.

src/use-cases/sync-latest-transcriptions.ts
aaronfay: Removed unnecessary `config` params

danielfay (approved):

PR #193, opened 2025-09-15
Add boilerplate for E2E playwright tests

danielfay:
Adds the basic setup for using playwright with Transcribe

Each dev will have to do a little bit of manual setup to be able to run the tests locally. There is a script to help with the setup that can be called with `npm run setup:e2e`. It will create a folder and an initial dotenv file if it doesn't already exist.

Login information needs to be added to the `.env` file, which we should probably store in 1Password. 

Supports an owner, editor and viewer user account for testing.

I've also added a helper for easily picking which account you would like to run a given test with `testWithAccount`. There are a few ways to achieve this in playwright, so hopefully this way works for us.

### Further Improvements to be Done
- decide which other browsers should be enabled
- github actions should `amplify push` and `amplify publish` when merged to `staging` and then run the e2e tests
- start adding meaningful tests
- create users for production
- update github actions to do deploys to production and then run e2e tests


aaronfay (approved):
Looks great, I like the structure of "all" vs just one type. 

## Further improvements priorities
 * GitHub Actions to run after `staging` deploy
 * Tests for permissions (view, edit, etc)
 * Critical functionality: upload transcription, peaks generate, spell checker works, add/delete region

danielfay:
Yeah, it turned out nice once all the accounts were required. The nice thing about having the separate accounts is that some tests will be only from the perspective of one type of user (for example, views have certain things they can't do that both editors and owners can). So, it'll be nice down the road when we right specific usecase tests :)

aaronfay:
I _suspect_ the bulk of the tests will be from the owner perspective, but definitely having "breakout" cases for different perms will be good. Nice work 👏🏻 👏🏻

PR #194, opened 2025-09-22
Githubactions deploy staging

danielfay:


danielfay:
That failure is finally more interesting, it is basically working, but we're now running into the issue of the teams json missing the proper references for the lambda layers.

I don't currently know how to fix that programmatically. Leaving this for now.

danielfay:
Closing stale draft for now, will approach this again at a later date

PR #195, opened 2025-09-29
Initial e2e tests

danielfay:
Adds tests for initial list mentioned in another PR

- Critical functionality: upload transcription, peaks generate, spell checker works, add/delete region

Transcriptions are created and cleaned up for each test.

e2e tests still need to be added for permissions. I still need to better understand how to use multiple different users in a test. An easier alternative may be to not have the test create the transcription and invite itself, but have that already exist and the users just login to check that they can view and edit it correctly.


aaronfay (approved):
Nice work 👏🏻

PR #196, opened 2025-10-06
Update `regionAnalysis` format to map

aaronfay:
This PR lays the groundwork for the language database, but focuses primarily on changing the `regionAnalysis` format from `string[]` to `WordMap[{word: '', analysis: '', allAnalysis: ['']}]`. This change is necessary for indexing later so we don't have to provide "second" analysis when indexing. In `allAnalysis` we store all possible analyses for that word so the user can select from the list (future feature).


src/services/migrationService.ts
danielfay, line 1: We can remove this after we have all of this deployed to production

src/use-cases/load-transcription.ts
```diff
@@ -99,8 +99,20 @@ export class LoadTranscription {
...
+            if (typeof item === 'object' && item.word) {
+              // Only add if analysis is complete (non-empty)
+              if (item.analysis && item.analysis !== '' && 
+                  item.allAnalysis && item.allAnalysis.length > 0) {
+                allKnownWords.add(item.word);
+              }
+              // Skip words with empty analysis - they need to be re-analyzed
+            } else if (typeof item === 'string') {
+              // Handle legacy string format
+              allKnownWords.add(item);
```
danielfay: This can go away

src/use-cases/subscribe-to-region-changes.ts
danielfay, line 1: Test that realtime changes are working with the new format

src/use-cases/update-region.ts
```diff
@@ -11,7 +12,7 @@ type RegionChanges = {
   translation?: string;
   start?: number;
   end?: number;
-  regionAnalysis?: string[];
+  regionAnalysis?: string[] | WordAnalysis[];
```
danielfay: We can remove `string[]` when this is deployed

src/use-cases/load-transcription.ts
aaronfay: Done.

src/use-cases/update-region.ts
aaronfay: Fixed

amplify/backend/function/crkSpellCheck/src/index.js
```diff
@@ -4,6 +4,27 @@ const PATH_TO_STRICT_ANALYZER = "/opt/crk-strict-analyzer.hfstol";
 const PATH_TO_RELAXED_ANALYZER = "/opt/crk-relaxed-analyzer.hfstol";
 const PATH_TO_STRICT_GENERATOR = "/opt/crk-strict-generator.hfstol";
 
+/**
+ * Converts macron diacritics to circumflex diacritics and strips punctuation
```
danielfay, line 8: I forgot we did this in this branch. I should be a bit more heavy handed with asking you to either break stuff out or link multiple issues to a PR.

We'll need some kind of agreement or strategy at least so that stuff like this makes it into the changelog on the docs site.

amplify/backend/function/crkSpellCheck/src/index.js
```diff
@@ -171,7 +238,10 @@ async function analyzeStrict(lookup) {
   
   for (const word of lookup) {
     try {
-      result[word] = fst.lookup(word);
+      const rawAnalyses = fst.lookup(word);
+      // Filter out error analyses (anything containing +Err/)
+      // These are fragments or orthographic errors that shouldn't be shown to users
+      result[word] = rawAnalyses.filter(analysis => !analysis.includes('+Err/'));
```
danielfay, line 244: Glancing below, to the relaxed version of this. Why don't we also need this filter on that one? It is working as is, so I don't think we need to change anything, but I'm surprised we don't filter the errors out of both raw analyses

amplify/backend/function/indexRegionData/src/lib/special-processing-crk.js
danielfay, line 1: Mostly a reminder to myself that this crk special processing should be reused by crkSpellCheck when that lambda gets refactored. That should happen shortly/when we add a second spellchecker.

amplify/backend/function/kiyanawlibHfstol/kiyanawlibHfstol-awscloudformation-template.json
danielfay, line 1: May we be so lucky that this is the last time this layer decides to update aha.

src/components/layout/AppLayout.tsx
danielfay, line 1: I know this is mostly just a pet peeve of mine, but I would prefer if you just deleted this instead of commenting it out. When you are done with the database stuff cursor will add the link back in without you noticing anyway, it's not like it will be forgotten.

If you really need an artifact to remind us of something that is left to be done it would be better to make an issue or a TODO instead of leaving commented out code blocks around the codebase.

src/components/player/WaveformPlayer.test.tsx
```diff
@@ -134,7 +134,7 @@ describe('WaveformPlayer', () => {
           selectedRegionId: null,
           selectedRegion: null,
           playbackWithinRegion: null,
-          knownWords: new Set<string>(),
```
danielfay, line 137: Oh, funny. I didn't know these were typed as `Set`s before. Technically the new type is duplication unsafe unlike the old one. Interesting.

src/hooks/useTextEditors.ts
```diff
@@ -163,13 +151,11 @@ export const useTextEditors = (regionId: string, activeTab: 'main' | 'translatio
...
-          text,
-          field: 'translation',
-          store: useEditorStore.getState(),
-          services
-        }).execute();
+        // Update store immediately for UI responsiveness
+        services.storeService.setRegionTranslation(regionId, text);
+
+        // Queue through save manager
+        regionSaveManager.queueTranslationChange(regionId, text);
```
danielfay, line 158: Seeing it in practice I really like this idea. Hopefully it will make saving more consistent going forward.

src/services/adt.ts
```diff
@@ -37,7 +37,7 @@ export interface RegionData {
   translation?: string;
   userLastUpdated?: string;
   index?: number;
-  regionAnalysis?: string[]; // Array of known words
+  regionAnalysis?: import('./spellCheckerService').WordAnalysis[]; // Array of detailed word analysis
```
danielfay: I think this should be the other way around. `adt` should define and export `WordAnalysis` and `spellCheckerService` should import it.

src/services/adt.ts
```diff
@@ -340,7 +388,7 @@ export class RegionModel {
   public translation: string;
   public userLastUpdated?: string;
   public index?: number;
-  public regionAnalysis: string[];
+  public regionAnalysis: import('./spellCheckerService').WordAnalysis[];
```
danielfay: Same as above comment.

src/services/index.ts
```diff
@@ -41,4 +42,7 @@ export const services = {
   conflictDetectionService,
   conflictResolutionService,
   versionConflictService,
+  regionSaveManager,
 };
+
+export { regionSaveManager } from './regionSaveManager';
```
danielfay, line 46: Does this need to export alone if it is also provided in services?

Or maybe the question the other way around is better. Why is a 'manager' in the services list?

src/services/regionSaveManager.ts
```diff
@@ -0,0 +1,542 @@
...
+    queue.lastChangeSource = 'bounds';
+    queue.lastChangeTime = Date.now();
+    
+    this.resetSaveTimer(regionId, 2500);
+  }
+  
+  /**
+   * Perform spell check and return analysis
+   */
+  private async performSpellCheck(regionId: string, text: string): Promise<WordAnalysis[]> {
```
danielfay, line 118: This feels like it belongs in the spellcheck service to me. It has nothing to do with queueing or saving.

src/services/regionSaveManager.ts
```diff
@@ -0,0 +1,542 @@
...
+        currentVersion
+      );
+      
+      // Update version after successful save
+      store.setRegionVersion(regionId, currentVersion + 1);
+      
+      // Update transcription metadata
+      const region = store.regionById(regionId);
+      if (region) {
+        const { UpdateTranscriptionUseCase } = await import('../use-cases/update-transcription');
```
danielfay: There a couple 'just in time' imports in this file. Can you please me them to the top of the file?

src/services/regionSaveManager.ts
```diff
@@ -0,0 +1,542 @@
...
+        // Determine primary field from the changes
+        const primaryField = changes.regionText !== undefined ? 'regionText' :
+                           changes.translation !== undefined ? 'translation' :
+                           'start';
+        
+        const store = services.storeService;
+        const currentVersion = store.getRegionVersion(regionId);
+        
+        // Handle the conflict with the dialog
+        await handleVersionConflict(
```
danielfay: This is a 109ln long function call. There much be a different for readable way to do this. There are async properties with ifs throughout. I'm having a really hard time understanding what is going on.

src/services/storeService.ts
```diff
@@ -11,15 +11,33 @@ import type { LazyRegion } from '../models';
 export const storeService = {
   // Editor Store operations
   // Region analysis operations
-  setRegionAnalysis: (regionId: string, analysis: string[]): void => {
+  setRegionAnalysis: (regionId: string, analysis: import('./spellCheckerService').WordAnalysis[]): void => {
```
danielfay: As noted earlier in the PR I would like to moved to `adt`, but beyond that, cursor was really intent on doing an inline import for this type eh? Very odd. It happens again later in this file.

src/services/versionConflictService.ts
```diff
@@ -17,7 +17,9 @@ type ErrorObject = {
 };
 
 // Type for region-like objects that may be incomplete
-type RegionLike = Partial<RegionData & LazyRegion>;
+type RegionLike = Partial<Omit<RegionData, 'regionAnalysis'> & Omit<LazyRegion, 'regionAnalysis'> & {
+  regionAnalysis?: import('./spellCheckerService').WordAnalysis[] | string | null;
```
danielfay: Here as well, interesting.

src/stores/useEditorStore.ts
```diff
@@ -39,8 +49,8 @@ interface EditorState {
   // Text selection state for creating issues
   regionSelections: Record<string, { index: number; length: number; text: string } | null>;
   
-  // Word analysis cache
-  knownWords: Set<string>;
+  // Word analysis cache - stores full WordAnalysis objects, not just strings
+  knownWords: Map<string, import('../services/spellCheckerService').WordAnalysis>;
```
danielfay: This is way better. It is even imported at the top of the file and still imported again inline here. Wild.

src/use-cases/load-transcription.ts
```diff
@@ -94,17 +94,40 @@ export class LoadTranscription {
   }
 
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  private extractKnownWordsFromRegions(regions: any[]): string[] {
-    const allKnownWords = new Set<string>();
+  private extractKnownWordsFromRegions(regions: any[]): import('../services/spellCheckerService').WordAnalysis[] {
```
danielfay: This import is truly the gift that keeps on giving. Really brightened up my morning ahaha.

danielfay (changes requested):

amplify/backend/function/crkSpellCheck/src/index.js
aaronfay: Yeah, this is still not fully "active" and though the changes should be _pretty close_ I'll wrap it up in the next branch. This branch actually just started out as "language database" but midway I realized we needed to refactor the analysis so it's a bit of a "pivot".

src/components/layout/AppLayout.tsx
aaronfay: Will do.

src/services/adt.ts
aaronfay: Good catch.

src/services/index.ts
aaronfay: Yeah good catch. I prefer it on its own at the end. I'll make that change.

src/services/regionSaveManager.ts
aaronfay: Yup good call. Moved this to the appropriate service file and also cleaned up some use cases that used to handle this.

src/services/regionSaveManager.ts
aaronfay: Fixed.

src/services/regionSaveManager.ts
aaronfay: I've busted this out into a bunch of smaller, named functions. Seems to be better now.

src/services/storeService.ts
aaronfay: Fixed.

src/services/versionConflictService.ts
aaronfay: Fixed!

src/stores/useEditorStore.ts
aaronfay: Yeah, bad cursor day I guess. Fixed.

src/use-cases/load-transcription.ts
aaronfay: Fixed.

amplify/backend/function/crkSpellCheck/src/index.js
aaronfay: Yup, good call.

danielfay (approved):
Double checked all comments. Looks good to me now, thanks!

danielfay (approved):
Great first pass on making the console more manageable. Thanks!

PR #197, opened 2025-10-15
Language database special processing

aaronfay:
Adds "special processing" case in the `indexRegion` Lambda so that language processing (indexing/analyzing) can be broken out. Also removes UofA (Sapir) Click-in-text API calls for analysis and just processes the provided analysis.


danielfay (approved):
Awesome, this will be great for adding new languages to process as well!

PR #200, opened 2025-11-10
Add transcription export

danielfay:
Screenshot 2025-11-10 at 4 11 30 AM

Adds a button in the transcription settings to export the transcription data to a text file. By default the file is in SRT format (with a `.txt` extension), but every element that isn't the 'Original text' can be toggled on or off before exporting


aaronfay (approved):
Looks great, nice work.

PR #204, opened 2025-11-21
Enable Language Database

aaronfay:
This PR adds a handful of changes to enable the Language Database into production:

 * fixes indexing so that deletes for a region are faster, rather than using a delete query (which scans the whole index) query for the documents for that region and bulk delete by ID
 * Don't allow viewers to export/download transcription data
 * Don't allow viewers to comment on issues unless "public issues" is enabled
 * tweak the stats page + pagination for verbs and nouns
 * deep link to attestations
 * fix PWA links to open in the same window
 * make searching more better
 * TOS opt-in warning users what the data can be used for.

screenshot-2025-11-20-17 53 17@2x

screenshot-2025-11-20-17 35 34@2x


amplify/backend/function/indexRegionData/src/lib/search.js
danielfay, line 1: I'll keep this pattern in mind for the future. Happy to hear that it enabled mass re-indexing

danielfay (approved):

PR #205, opened 2025-11-21
Issue UI Improvements

aaronfay:
PR adds a couple enhancements to the UI for usability:

 * ability to toggle on/off which issues types to view
 * ability to search issues by text
 * adds next/previous buttons to issue list, "previous" at Region 1 will navigate to last region

screenshot-2025-11-21-14 06 11@2x

screenshot-2025-11-21-14 06 18@2x

### DIFF
--- a/src/components/issues/IssuesPanel.tsx
+++ b/src/components/issues/IssuesPanel.tsx
@@ -10,6 +10,7 @@ import { useNavigateIssueRegions } from '../../hooks/useNavigateIssueRegions';
 import { useIssueFlashIndicator } from '../../hooks/useIssueFlashIndicator';
 import { FLASH_CONFIG } from '../../services/flashIndicatorService';
 import { useSelectAndPlayRegion } from '../../hooks/useSelectAndPlayRegion';
+import { type IssueType } from '../../services/adt';
 
 // Suggestion Popover Component
 interface SuggestionPopoverProps {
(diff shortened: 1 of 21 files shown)


src/components/issues/IssuesPanel.tsx
```diff
@@ -404,6 +404,11 @@ interface IssuesPanelProps {
   onDeleteIssue: (issueId: string) => void;
   variant?: 'modal' | 'bottom-sheet';
   onJumpToRegion?: (regionId: string) => void;
+  enabledIssueTypes?: Set<'needs-help' | 'indexing' | 'new-word'>;
```
danielfay: This is all over so I'm only noting it on the first one.

https://github.com/search?q=repo%3Akiyanaw%2Fkiyanaw-backend%20%27needs-help%27%20%7C%20%27indexing%27%20%7C%20%27new-word%27&type=code

It looks like we do have an explicit type here: https://github.com/kiyanaw/kiyanaw-backend/blob/6331af4d5e1d827bb2b650d8445077bb869ca996/src/services/textHighlightService.ts#L9

But that may not be the best place for it. It should probably be moved to `adt` and then used everywhere.

The typing seems consistent across all these files so it would be a good idea to have it live in one place.

danielfay (approved):
Looks good, up to you if you want to do the clean up right now.

PR #217, opened 2025-11-25
Generic spellcheck lambda backed with EFS

danielfay:
Creates a new lambda function for spellchecking that is backed by an EFS to support varying file sizes of FST.

- Uses new generic /bulk-lookup and /suggest endpoints with `languageCode` in the body of the POST instead
- Downloads HFSTOLs to a mounted EFS instead of relying on a lambda layer
- Code broken out of `index.js` and with tests

### Future improvements

- EFS and access point creation is manual just like the peaks lambda. Would be nice to have it done automatically
- Same with connecting the VPC and EFS to the lambda itself
- The old language specific lambdas and layer can be removed after this is released
- This PR does not include any new language support, only a redesign of the existing spellchecking


aaronfay (approved):
Looks good. I can see us fleshing this out down the road if we end up with a lot of FSTs we need to support, but this is a good start 👍🏻

PR #219, opened 2025-11-25
Analysis picker, spelling suggestions

aaronfay:
A few refinements for the RTE:

 * moves the tabs for Original / Translation into toggles on the button bar
 * utilizes that space for a "region bar" which is a context bar
 * shows blue underline when analyzed word has multiple (ambiguous) analyses
 * shows grey underline when spelling suggestion is available
 * region bar shows available options for both
 * adds "Database" button that links to word form in the Language DB when under the cursor

Also tweaked the issues panel to be resizable and a few other CSS tweaks.

screenshot-2025-12-30-16 44 21@2x
screenshot-2025-12-30-16 45 49@2x


src/components/regions/RegionEditor.tsx
```diff
@@ -25,13 +27,33 @@ export const RegionEditor = memo(({
...
 
+  // Get word under cursor for dictionary lookup
+  const { cursorWordAnalysis } = useRegionContextBar(region.id, canEdit);
+  
+  // Extract lemma from analysis, skipping prefixes like PV/, IC+, RdplW+, RdplS+
+  // e.g., "PV/e+PV/ki+ohcîw+V+AI+Cnj+3Sg" -> "ohcîw"
+  const extractLemma = (analysis: string): string | null => {
+    if (!analysis) return null;
+    const parts = analysis.split('+');
+    const prefixes = ['PV/', 'IC', 'RdplW', 'RdplS'];
```
aaronfay: I just realized this is naive, there will be more than this. I need to think about how to approach the lemma here if we are dealing with different FST analysis patterns/tagging. 🤔

amplify/backend/function/crkSpellCheck/src/index.js
danielfay, line 1: This whole function can be removed if you like. You can also leave it and I'll remove it next time I'm cleaning things up.

danielfay (approved):
Lot's of small changes in this that I really like.

Hopefully having the shared language processing will help keep efforts between the indexer and app in sync.

There is quite a bit of complicated and winding logic in the rte service changes. As long as it is working I'm happy enough with it. This PR added about 800 lines to that file though, feels like it could be an app in itself aha.

src/components/regions/RegionEditor.tsx
danielfay: After our meeting we decided to make the tag processing (and general index related special processing) a shared library that the frontend can use as well.

PR #220, opened 2025-11-25
Add support for otwc and otwr

### LINKED ISSUE #206: Add spell-check support for Odawa
After our meeting recently with Dustin, add support for the additional FST for anishnaabemowin `otw` (Odawa). This looks to be compatible with the `hfst` tooling we currently using (being based on GeillaLT) so we should be able to reuse some of the same layers to make it happen. 

Make it so! 👉🏻

danielfay:
Adds support for otw (otwr and otwc - naming ours) with hfstol files provided from https://github.com/bowersd/otw/releases

Because the hfstol library we are currently using doesn't support weighted transducers the strict and relaxed analyzers are currently not taken from the release, but added to https://github.com/bowersd/otw/blob/master/src/hfst-compilation.sh using the `-O` instead of `-w` flag in the `hfst-fst2fst` steps.

To support further languages the language selector has also been adjusted to be searchable.

hfstol files have already been uploaded to both staging and production

### Testing

Used example sentences from https://bowersd.github.io/textAnalysis/sentences.html to test each of the new FSTs and got some of the words to analyze


src/components/forms/SearchableLanguageSelector.tsx
```diff
@@ -0,0 +1,163 @@
...
+                <button
+                  key={language.code}
+                  type="button"
+                  onClick={() => handleSelect(language.code)}
+                  className={`
+                    w-full px-4 py-2 text-left hover:bg-gray-50 focus:bg-gray-50 focus:outline-none
+                    ${value === language.code ? 'bg-blue-50 text-blue-900' : 'text-gray-900'}
+                  `}
+                >
+                  {language.name}
```
aaronfay, line 149: Just one tweak here I been thinking about, can we have the language display as `Plains Cree Y-dialect (crk)` as well? This will correspond to some changes I have planned for the documentation where we can list off some of the unique "idiosyncrasies" (is that the right word?) or details about various FSTs because they may not all necessarily behave exactly the same?

aaronfay (approved):
I'd love to just see that one little tweak, otherwise looks awesome 👍🏻

src/components/forms/SearchableLanguageSelector.tsx
danielfay: I can do that, though I have other thoughts on that language naming as well.

The language list currently looks like:

```
  { code: 'crk', name: 'Plains Cree Y-dialect' },
  { code: 'crgn', name: 'Northern Michif' },
  { code: 'otwc', name: 'Nishnaabemwin (Corbiere-style)' },
  { code: 'otwr', name: 'Nishnaabemwin (Rhodes-style)' },
```

I would kinda prefer them to all be similar in naming style. I can't make up my mind though. The idealist in me would want the languages to be 'in the language' like the bottom 2. So, Nēhiyawēwin for Plains Cree right?

Alternatively I can switch the bottom two to be 'Eastern Ojibwe' instead of Nishnaabemwin.

What are your thoughts?

aaronfay: Yeah actually `nêhiyawêwin (Plains Cree Y-dialect)` is how I usually write it, but that should be an easy change. Michif is just "Northern Michif" 🤷🏻 

I wanted the language code in there but actually it might not matter

aaronfay: Actually, forget the language code but can we do the following:

```
nêhiyawêwin (Plains Cree Y-dialect)
Michif (Northern)
Nishnaabemowin (Odawa - Corbiere)
Nishnaabemowin (Odawa - Rhodes)
```

And then `ciw` would be:

```
Anishnaabemowin (Ojibwe)
```

PR #221, opened 2025-12-05
force fresh AWS credentials before signed URL

aaronfay:
Often when transcribing we'll get the "waveform error" repeatedly caused by a 403 from S3. In those instances it _appears_ that Cognito credentials can be old and run out, and will fetch a few range requests but then error out again. 

This change forces fresh Cognito credentials which should give us consistently 1 hour of transcribing before the error. It's still a shot in the dark, the issue is hard to reproduce consistently but this is my first attempt to try and correct it. I believe using a CDN for the media files will also be helpful, but I'll set that up in another PR.


src/hooks/useSelectAndPlayRegion.ts
```diff
@@ -1,12 +1,37 @@
-import { useCallback } from 'react';
+import { useCallback, useRef } from 'react';
 import { SelectAndPlayRegion } from '../use-cases/select-and-play-region';
 import { services } from '../services';
 import { useEditorStore } from '../stores/useEditorStore';
 
 export const useSelectAndPlayRegion = () => {
-  // not debounced so we can call multiple times
+  // Add debouncing to prevent rapid region switching that can cause race conditions
```
danielfay: Is this directly related to the attempted fix for the waveform error or does this fix something else?

I'm fine with it being in here, but patch notes wise if it's meaningful then we may as well list it.

src/services/transcriptionService.ts
```diff
@@ -118,6 +122,55 @@ export const generateSignedUrl = async (sourceUrl: string, fileSuffix: string =
...
 };
 
+// Track when credentials were last refreshed
+let credentialSetupTime: number | null = null;
+
+/**
+ * Forces fresh credentials to be obtained.
+ * Call this when loading a transcription to ensure we have a full hour of valid credentials.
+ */
+export async function forceFreshCredentials(): Promise<void> {
```
danielfay, line 129: I like it being encapsulated like this.

It means that if it's useful down the road and we can better figure out what type of error we're getting during the waveform error (like maybe one is a 404 and the other a 5XX), we can call this to get new creds and retry.

src/use-cases/load-transcription.ts
```diff
@@ -30,6 +30,11 @@ export class LoadTranscription {
     const spellCheckerService = this.config.services.spellCheckerService
     const userService = this.config.services.userService
     
+    // Force fresh credentials when loading a transcription to ensure we have a full hour
+    // Import forceFreshCredentials from transcriptionService module
+    const transcriptionServiceModule = await import('../services/transcriptionService');
```
danielfay: Is this really not in `this.config.services`?

src/services/transcriptionService.ts
danielfay: Back from reading all of the PR, and it looks like you are doing the logging to see if the above is possible. Awesome 🎉

danielfay (approved):

src/use-cases/load-transcription.ts
aaronfay: Oops, missed that one 🤦🏻

src/hooks/useSelectAndPlayRegion.ts
aaronfay: You're right. I think I'm going go pull the right out, not exactly sure what problem it's solving now.

PR #223, opened 2025-12-09
Add media expiry timer + dialog

aaronfay:
Further attempting to improve the UX around expiring media credentials, this PR adds in a timer that will show the remaining time left for the current credentials. Clicking it will give the option to refresh, and it will also automatically pop up once the credentials have expired, prompting the user to refresh.

screenshot-2025-12-08-16 10 55@2x

screenshot-2025-12-08-16 12 18@2x


danielfay (approved):
Looks great!

PR #224, opened 2025-12-15
Add language ciw

danielfay:
Adds support for `ciw` Anishnaabemowin (Ojibwe) with a strict transducer created from https://github.com/ELF-Lab/OjibweMorph

Also made the lambda hfstol downloading check to see if there is a new version of the file in S3 and download it if so, instead of never looking if it had the hfstol on disk already

This hfstol is already uploaded to staging and production

### Testing

Tested with a couple words from the hfst itself. Text was analyzed and returned as a known word


aaronfay (approved):
Looks great 🎉

PR #226, opened 2025-12-28
Add indexing support for `ciw` and `otw`

aaronfay:
Adds in the appropriate special processing files to handle `ciw`, `otwr` and `otwc`.


amplify/backend/function/indexRegionData/src/lib/special-processing-ciw.js
```diff
@@ -0,0 +1,165 @@
...
+    }
+  }
+
+  // Split the analysis string by '+'
+  const parts = analysis.split('+')
+  
+  // Find the lemma - first part that doesn't start with a special prefix
+  let lemma = null
+  let lemmaIndex = -1
+  const prefixes = ['PVDir/', 'PVTense/', 'PVLex/', 'PVSub/', 'PVRel/', 'ChCnj', 'PNLex/']
```
danielfay, line 76: Interesting, I notice that this list is different for each of the 3 files. How did you go about generating it?

amplify/backend/function/indexRegionData/src/lib/special-processing-crk.test.js
danielfay, line 1: I originally had thought that maybe this was a mistake and you had meant to move this into the `test` folder, but looking at the files in there one already exists. Good clean up!

danielfay (approved):
Looks good! Just a question so that I understand you process when we add more languages.

amplify/backend/function/indexRegionData/src/lib/special-processing-ciw.js
aaronfay: Dumped a huge number of examples into Claude and said "make tests for these" 🤷🏻

aaronfay: From here specifically: https://github.com/ELF-Lab/OjibweMorph/tree/main/scripts/test_data

I _assume_ when folks are building out FSTs they generally have a bunch of tests to make sure stuff is coming out as expected, so that'll probably be the best place to run to moving forward to get good sample data for the "tag" variations an FST might need to handle.

PR #227, opened 2025-12-29
Another stab at token expiry

aaronfay:
I _believe_ what's happening is when we prompt the user to "refresh media token" because it has expired, we don't _actually_ get new credentials, so the problem persists. This skips "window.refresh()" and actually just refreshes everything in-place, streamlining the UX a little.

If this goes well, I will probably turn this into an "auto-refresh" because it's automatically getting a new URL and actually updating wavesurfer without reloading the page, that would be an obvious next step.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -63,3 +63,30 @@ playwright-report/
 test-results/
 .env
 debug-*.png
+
+# Claude Flow generated files
+.claude/settings.local.json
+.mcp.json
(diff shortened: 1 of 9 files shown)


danielfay (approved):
A bit frustrating that the other try didn't fix it.

This looks good, if it fixes it then I totally agree that making the new URL behind the scenes is the way to go next

PR #229, opened 2026-01-14
IaC remaining backend wiring

### LINKED ISSUE #216: Deployment script
Would be great to have a deployment script eg: `npm run deploy-prod` that can

 * run `amplify publish --yes` for the given env
 * double-check the `peaks` and `spellcheck` Lambda VPC and related EFS are still attached
 * runs an invalidation against Cloudfront for the env

Make it so! 👉🏻

danielfay:
This PR mostly adds amplify post push and publish hooks that use serverless and some scripts to do addition backend wiring. Automatic Cloudfront invalidation on publish is also handled now.

Some new lock files came along for the ride when I got re-setup so there are a lot of lines changed.

A function was also removed from `src/services/rteService.ts` because it apparently wasn't being used and was causing publishing to throw an error.

### Next Steps

To deploy this to production the manual wiring that it codifies needs to be deleted, this includes:
- Delete the VPC endpoint to S3
- Unhook the VPC and EFS from the peaks lambda, then delete the EFS and it's access point
- Unhook the VPC and EFS from the spellcheck lambda, then delete the EFS and it's access point

### Testing

- Successfully ran both `amplify push` and `amplify publish` in staging and verified that the deleted wiring was back
- Created a new transcription
- Spellchecked some words


aaronfay (approved):
AND it invalidates CF distribution 🎉  nice work 👍🏻

PR #232, opened 2026-02-23
CDN for streaming media

### LINKED ISSUE #222: CDN for streaming media
Rather than streaming the media from S3 (which can be chunky) add a CDN into the mix to help reduce latency in streaming media from the cloud.

URLs for the CDN should be signed and valid for 24 hours.

danielfay:
closes #222 

This PR puts a cloudfront in front of the transcription S3 bucket. Signed URLs for cloudfront work a little different than the ones to S3 directly. They need to be signed with a key pair. The public keys are in the serverless directly of the repo, while the private keys have been added as AWS secrets.

A new API and lambda have been created that return signed URLs for requested resources in the CDN. The frontend now calls that lambda instead of creating it's own URLs. Uploads are still done directly to S3.

### Testing

Tested on my staging account by loading a few old transcriptions and uploading a couple new ones. The network tab shows usage of the new CDN URLs


aaronfay (approved):
Very cool, nice work 👍🏻

PR #235, opened 2026-03-07
Fix CDN 403

danielfay:
Fixes an encoding issue with the new CDN.

S3 keys with spaces weren't properly being encoded before the URL was signed, causing a mismatch when the signed URL was given back to the frontend.

A little bit of cleanup also snuck into this PR. My staging amplify was in an odd broken state based on some failed deletions I did in a different branch. The unneeded old routes in the spellcheck API have been removed, which allowed me to push.

### Testing

Published to staging, clicked through a variety of my transcriptions and they are all loading.


aaronfay (approved):
Good catch 👌🏻

Issue #230, opened 2026-01-14
Add Siksika (Blackfoot) language support


danielfay:
Repo found here: https://github.com/giellalt/lang-bla

danielfay:
I have this repo building locally. While not tested, I assume any hfstols generated are weighted so on hold for now.

PR #238, opened 2026-03-23
Added support for BLA, CWD and GLE

### LINKED ISSUE #215: Add support for TH transducer
Repo is here: https://github.com/giellalt/lang-cwd

danielfay:
resolve #215 #230 #231 

Add 3 new languages to the language drop down when configuring a new transcription. Indexing of the public language information to the language database will be done at a later date.

Also, deletes the old spellcheck API Gateway, which is no longer needed and was in a semi-bricked state in staging.

### Testing

Verified that new languages were in the drop down on staging and that when using the new languages the proper language codes are used.


aaronfay:
Need testing on this? As soon as this rolls out I'll make some videos and announcements and stuff. lmk

aaronfay (approved):

PR #240, opened 2026-03-27
Indexing for BLA, CWD and GLE

### LINKED ISSUE #239: Add language database indexing for BLA, CWD and GLE
While these languages are available to users, special processing needs to be added so that public transcriptions can be added to the language database.

danielfay:
Adds special processing needed to prep lemmas for being added to the language database for BLA, CWD and GLE

For GiellaLT repos (which the 3 above are) the fst test ymls were used for tag examples that can be parsed to find the lemma for a given word

### Testing

Made a public transcription on staging and saved some words to the language database


aaronfay (approved):
Beautiful 🚀

PR #243, opened 2026-04-13
Integration tests for some common usecases

### LINKED ISSUE #242: Expand integration tests for common user usecases
The currently playwright tests were only to get the framework in place and as a proof of concept.

Before any sweeping project changes are made they should be built out into a proper starting suite of common use cases.

danielfay:
Before doing a security audit some more integration tests seemed like a good idea. These tests a decent (but not complete) amount of the basic functionality in transcribe.

One notable change that is on the edge of the scope of this issue is an update to the invite lambda. Invites that fail to send their email are no longer set to `failed`, but are instead left as pending and can still be accepted from the invite page if the user can login.

### Testing

- Tests and linting passed.
- New integration tests ran against staging and pass


aaronfay (approved):
Nice work. I can appreciate some of those were probably tricky to get working, but the more coverage the better. 👏🏻 👏🏻

PR #250, opened 2026-05-22
Cleanup waveform processing

danielfay:
Improves how waveform peaks are generated and processed to reduce the risk of sync issues with non-standard audio files, and adds tooling for users to recover from bad peaks files.

Also upgrades WaveSurfer from 7.9.5 to 7.12.7.

Note: waveform desync is not fully resolved by this PR. A future project to transcode all uploads to a standard format at upload time is the intended permanent fix.
  
## Testing

Updated unit tests across the Lambda, wavesurfer service, transcription service, and load-transcription. Manually tested the regenerate button with a problematic MP3.


aaronfay (approved):
👍🏻

PR #252, opened 2026-05-23
Add deployment slack notifications

danielfay:
Adds Slack notification support for both push and publish deployments. Setting up your own Slack app and providing a webhook URL enable the notifications.

### Testing

Did both new pushes and publishes from this branch and saw notifications in our builds channel
Tested a broken deploy to see failure notification


aaronfay (approved):
Awesome, thanks

PR #254, opened 2026-05-29
Fix region autoplay after wavesurfer update

danielfay:
Fixes region playback needing two clicks to actually start playing. Also fixes an unrelated region bug where the region play button didn't always respect that I new region had been selected (specifically when selecting it in the waveform).

If I had to guess when I updated the wavesurfer package it slightly changed the behavior.

### Testing

Added regression tests for the bugs that first failed and now pass
Deployed to staging and verified that region playback works in one click again.


aaronfay (approved):
Working good, thanks for this.

PR #256, opened 2026-06-02
Remove spellcheck infrastructure artifacts

danielfay:
Removes the Docker/docker-compose setup and EFS infrastructure that were used by the old Node.js spellcheck lambda, which has since been replaced. Cleans up the Dockerfile, docker-compose.yml, the spellcheck EFS/mount target/access point resources from serverless.yml, the VPC attachment script entry, and the related README sections.

### Testing

No automated tests for infrastructure teardown. Verified by diffing that all spellcheck EFS references are gone from serverless.yml and the attach script.


aaronfay (approved):
Less code +1

PR #261, opened 2026-06-03
Add spacebar play/pause shortcut

### LINKED ISSUE #53: space bar not stopping audio when not in edit mode
While playing the audio and not editing (not with an active region) space bar doesn't stop the audio

danielfay:
resolve #53

Adds a global `keydown` listener on `WaveformPlayer` that toggles play/pause on spacebar. The handler is guarded to no-op when focus is inside an `input`, `textarea`, or `contenteditable` element, so it doesn't interfere with typing in region text fields.

### Testing
Unit tests in `WaveformPlayer.test.tsx` cover the play, pause, and focus-guard cases. Manually verified spacebar toggles playback and does not fire while typing in a region.


aaronfay (approved):
Good one.

PR #264, opened 2026-06-05
Fix invite case sensitivity, expiry reset, and permission changes

danielfay:
resolves #263

Invites had three problems: case-sensitive email matching caused 500 errors when the stored address didn't exactly match the authenticated user's email; expired invites had no recovery path; and collaborator permission levels couldn't be changed after the initial invite. This adds email normalisation on both send and accept, a renew/delete flow for expired invites, and a permission-update action accessible from the transcription settings page.

A migration script (normalize-invite-emails.js) is included to lowercase existing invite records.

### Testing
Unit tests cover the invite service, renew-invite, send-invite, and update-invite-permission use-cases on both the Lambda and frontend sides. E2E tests in invite-management.spec.ts cover the new flows. Manually verified invite accept with mismatched email case, renewing an expired invite, and changing a collaborator from viewer to editor.

### DIFF
--- a/amplify/backend/api/APIGatewayAuthStack.json
+++ b/amplify/backend/api/APIGatewayAuthStack.json
@@ -298,6 +298,126 @@
                       },
                       ":",
                       {
+                        "Ref": "invite"
+                      },
+                      "/",
+                      {
+                        "Fn::If": [
+                          "ShouldNotCreateEnvResources",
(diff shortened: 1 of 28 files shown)


aaronfay (approved):
Nice work, I was fixing these manually. 👍🏻

PR #267, opened 2026-06-05
Fix waveform scrollbar clipping into controls

danielfay:
resolve #266

Changed the outer media area from `h-32` to `h-40` (160px), which properly contains WaveSurfer's full 148px rendered height with a 12px buffer for the scrollbar, and removed the `mt-[20px]` workaround.

### Testing

Existing WaveformPlayer tests pass. Manually verified the scrollbar no longer clips into the controls when the waveform is zoomed in.


aaronfay (approved):
Nice one.

PR #272, opened 2026-06-05
Add processMedia Lambda for media transcoding

danielfay:
resolve #269

Adds a processMedia Lambda triggered by the MediaTable DynamoDB stream. Transcodes incoming media to a consistent MP4 or MP3, generates waveform peaks, and produces a thumbnail for video. The invite handler now syncs Media ACLs alongside Transcription ACLs when sharing is accepted or revoked.

Note: the upload flow that creates the initial PENDING Media record is covered in #270, where primary end-to-end testing will take place.

### Testing

230 unit tests pass. Deployed to staging without error.


aaronfay (approved):
Seems fine. If we have this on staging, I'd want to test it out with a number of different files/formats specifically to make sure that the transcoding goes as expected. even throw some old .wmv files at it or something to see. Also specifically the 48K mp3 that Allison was having issues with, make sure those work too (if we haven't tested that already)

PR #273, opened 2026-06-06
Fix "shared with me" tab showing stale transcriptions after logout or account switch

danielfay:
resolve #265

Clears the transcription service cache and resets `useTranscriptionsStore` and `useInviteStore` before signing out in `AppLayout`, so the next login always starts with a clean slate. Also handles the back-to-back account switch case in `AuthenticatorShell`, where one user logs out and another immediately logs in within the same session without a page reload.

### Testing
New unit tests in `useTranscriptionsStore.test.ts` and `useInviteStore.test.ts` verify that `reset()` clears all state fields. The second store test explicitly covers the account-switch scenario, asserting that the previous user's transcriptions are not visible to the incoming user. Manually tested locally by switching between two accounts and confirming the "shared with me" tab shows only the current user's data.


aaronfay (approved):
Good catch

PR #287, opened 2026-06-08
Fix overlapping region auto-selection during bounded playback

danielfay:
resolve #286

When two regions overlap in time, playing back the first region caused the wavesurfer `region-in` event to fire for the overlapping region too, jumping both the editor selection and the waveform highlight to the wrong region. Fixed by gating the service-level and hook-level `region-in` handlers to ignore events for any region other than the one playback is currently bounded to.

### Testing
Three new tests cover the fix: the service handler skips waveform tinting for overlapping regions during bounded playback, and the hook handler skips selection/highlight/scroll for non-bound regions while correctly proceeding for the bound region. The `getPlaybackBoundRegion()` getter is also tested directly. Full suite (2069 tests) passes. Also verified locally with two overlapping regions — selection, waveform highlight, and typing all stay on the played region throughout playback.


aaronfay (approved):
One line fix. _Many tests_ 😂  nice work 👍🏻

PR #293, opened 2026-06-09
Wire upload flow and show media processing status

danielfay:
resolve #270
resolve #271

Creates a Media record alongside each new transcription so the processMedia Lambda can transcode it. The transcription list and editor now show processing status while the media is being prepared, polling until it's ready. Deleting a transcription cascades to its Media record and S3 files. Also adds a delete button to the transcription list for owners.

### Testing
Unit tests cover the new mediaService, uploadService, and the create/delete transcription use cases. Manually verified upload creates a Media record, the processing indicator appears and clears when transcoding finishes, and deletion cleans up both the record and its S3 files.

### DIFF
--- a/amplify/backend/api/kiyanaw/schema.graphql
+++ b/amplify/backend/api/kiyanaw/schema.graphql
@@ -179,6 +179,7 @@ type Media
   status: String!
   originalKey: String!
   renditionKey: String
+  peaksKey: String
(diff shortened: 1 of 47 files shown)


aaronfay (commented):
Code looks fine, are we going to auto-migrate the existing transcriptions? Or do they just load retroactively?

PR #311, opened 2026-07-14
Fix shared transcription load failure blocking entire transcription list

danielfay:
resolve #303

A shared-transcription load failure (e.g. a stale PWA bundle calling an unregistered REST API, which surfaced as "API name is invalid") was propagating through `performSync` and blanking the entire transcription list. The shared-load call is now isolated in its own try/catch so a failure degrades gracefully — owned transcriptions still render, and the error is logged with build hash and service worker URL to help identify stale-bundle occurrences in production.

### Testing
- Added unit tests in `transcriptionService.test.ts` covering the shared-load failure path.
- `npm run test` (1993 tests) and `npm run lint` pass.

### DIFF
--- a/src/services/transcriptionService.test.ts
+++ b/src/services/transcriptionService.test.ts
@@ -926,6 +926,48 @@ describe('TranscriptionService', () => {
     });
   });
 
+  describe('performSync resilience — shared load failures (issue #303)', () => {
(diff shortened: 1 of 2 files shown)


aaronfay (approved):
lgtm 👍🏻

